PR #7, opened 2020-06-10
Improved bindCredentials example and added info in usernameKey

whazenberg:


mpvvliet:
LGTM. Be aware @whazenberg that this needs to be merged to the v4.0.1 branch before it is live.

PR #34, opened 2020-08-26
Added clarification about read permissions

hruhek:


integrations/available_stackpacks/nagios.md
```diff
@@ -49,11 +49,14 @@ instances:
...
-sudo /etc/init.d/stackstate-agent restart
+sudo service stackstate-agent restart
 ```
 
-Once the API-Integration Agent is restarted, wait for the Agent to collect the data and send it to StackState.
+Once the StackState Agent is restarted, wait for the Agent to collect the data and send it to StackState.
 
+## Permissions for Nagios files
+
+Nagios StackState Agent check tails Nagios config and log files so it should have permission to read those files. If you run StackState Agent with some other user than `root`, you need to provide a way for StackState Agent check read those files. Sometimes these files can be recreated by Nagios, so manually setting read permission is not an option. You should add StackState Agent user to same group witch is attached to those log files.
```
gggina, line 62: Nagios StackState Agent check tails Nagios config and log files, so it should have permission to read those files. If you run StackState Agent with some other user than `root`, the StackState Agent user must be added to the same group that is attached to the config and log files. Note that manually setting read permission is not an option as the files can sometimes be recreated by Nagios.

hruhek: Fixed.

gggina (approved):
looks good

PR #36, opened 2020-08-28
Stac 9832 how to setup traces

gggina:
I placed this under 'configure' - happy to move it if you think it sits better somewhere else


mpvvliet (approved):

mpvvliet:
Would be nice if we can include the installation instructions for the tracing integrations also in the docs site. They are now part of the stackpack in this repo: https://gitlab.com/stackvista/integrations/stackpacks. 

For example: https://gitlab.com/stackvista/integrations/stackpacks/-/blob/master/stackpacks/stackstate-agent-v2/src/main/stackpack/resources/int-java-overview.md

PR #38, opened 2020-08-31
STAC-7537: [4.2] Drop obsolete reindex documentation

craffit:
This functionality will be dropped, will only be part of 4.2


gggina:
Also remove the below article from support KB when published:

**Error `java.lang.IllegalStateException: Requested index specs do not match the catalog.**

PR #39, opened 2020-09-03
Stac 10363 kubernetes log file info

gggina:
first review.


configure/stackstate_log_files/kubernetes_sts_log_files.md
```diff
@@ -0,0 +1,60 @@
...
+| Data ingestion | `stackstate-receiver` |
+| Event handlers | `stackstate-server` |
+| State propagation | `stackstate-server` |
+| Synchronization | `stackstate-server` |
+| View health state | `stackstate-server` |
+
+You can access logs on a specific pod using the [`kubectl logs` command](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#logs). For example:
+
+```
+kubectl logs stackstate-server
```
rb3ckers, line 48: I just realized the pod-names I've given are only prefixes. Depending on the type of the application they either have a number-suffix (like `stackstate-server-0`) or a random string (like `stackstate-receiver-5b9d79db86-h2hkz`).

To make things worse even the `stackstate` prefix is dynamic. If customers would deviate from the standard installation flow and use a different release name than the suggested `stackstate`  in the `helm upgrade` command the pod name would change to something like `<release-name>-stackstate-server-0`. This still contains `stackstate-server` though. So I'm not sure if we should explain all this detail in the docs, it may confuse our users more then help them.

configure/stackstate_log_files/kubernetes_sts_log_files.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+{% hint style="danger" %}
+Note that logs stored on pods will be regularly removed. For long term access to logs, it is advised that you set up [log aggregation](#log-aggregation) for your Kubernetes cluster.
+{% endhint %}
+
+
+## Log aggregation
+
+For long term storage of StackState log data, it is advised that you set up log aggregation on your Kubernetes cluster. This can be done using a third party system, such as Elasticsearch, Splunk or logz.io
```
rb3ckers, line 58: "By using a third party system for storage such as Elasticsearch, Splunk or Logz.io and a log shipper such as Logstash or Fluentd".

This is something you install for your whole kubernetes cluster typically. Installing this is a whole docs page (or more) on its own and depends on the tools. So I'd refer to official docs and/or blog posts (the latter is usually a more complete overview but more likely to become incorrect:
* Fluentd docs for shipping logs: https://docs.fluentd.org/container-deployment/kubernetes
* A complete overview of setting up log aggregation into Elasticsearch: https://docs.bitnami.com/tutorials/integrate-logging-kubernetes-kibana-elasticsearch-fluentd/

rb3ckers (approved):

PR #73, opened 2020-09-30
Stac 10 new nav configure

gggina:
This is a quickfix to clean up the configure section:
- 'configure' items grouped alphabetically under topic
- log files now under 'setup'
- 'use' section moved above 'stackpacks'


mpvvliet (approved):
Looks good!

PR #97, opened 2020-10-13
Added upgrade note for new events perspective

Q69K:


setup/upgrading.md
```diff
@@ -117,6 +117,18 @@ Once StackState has been upgraded and started, verify that the new installation
 
 ## Version-specific upgrade instructions
 
+### Upgrade to 4.2.0
+
+* The Events Perspective has been updated to include more comprehensive events.<br />Important notes:
+  * Events submitted before upgrade will not be available in the Events Perspective after upgrade.
+  * Streams using events will continue to work as before.
+  * Deprecated: 
+    * The `sts_generic_events-*` Elastic index has been deprecated in favor of `sts_topology_events-*`.
```
craffit, line 126: We are not ready to deprecate this just yet.. we have no alternative to offer. We should first have that and then start deprecating

setup/upgrading.md
```diff
@@ -117,6 +117,18 @@ Once StackState has been upgraded and started, verify that the new installation
 
 ## Version-specific upgrade instructions
 
+### Upgrade to 4.2.0
+
+* The Events Perspective has been updated to include more comprehensive events.<br />Important notes:
+  * Events submitted before upgrade will not be available in the Events Perspective after upgrade.
```
craffit, line 123: This is more like an release note, the upgrade notes should detail actionable steps that need to be taken to get the upgrade done

setup/upgrading.md
```diff
@@ -117,6 +117,18 @@ Once StackState has been upgraded and started, verify that the new installation
 
 ## Version-specific upgrade instructions
 
+### Upgrade to 4.2.0
+
+* The Events Perspective has been updated to include more comprehensive events.<br />Important notes:
```
craffit, line 122: I am missing the one actionable step that we make changes to the processmanager.conf by added the topologyEvents2ES service. this is what some customers may need to fix if they have a customized config

craffit (changes requested):

setup/upgrading.md
Q69K: Good point! Will check for another files

PR #111, opened 2020-10-21
Stac 10 logging

gggina:
Moves logging to the 'configure' section and adds details to set logging level and include logging messages in functions


configure/logging/logging-for-functions.md
```diff
@@ -0,0 +1,240 @@
...
+    "id": 154786643410143,
+    "name": "test",
+    "type": "ComponentActionDefinition"
+  }
+]
+```
+{% endtab %}
+{% endtabs %}
+
+#### Event handler functions
```
mpvvliet, line 84: This is kind of tricky. The event handler functions are debugged on an **event handler** level. So, you might have view A and view B that both have the Slack event handler function configured. You can enable logging on only one of the event handlers (in view A or view B), but not on the function itself (which would show logging if the event handler in either view would go off).

configure/logging/logging-for-functions.md
```diff
@@ -0,0 +1,240 @@
...
+
+- In the [StackState UI Analytics environment](/develop/scripting/README.md#running-scripts) enter the query below to list all event handler functions.
```
mpvvliet, line 86: The query below indeed retrieves event handlers, not event handler functions.

configure/logging/logging-for-functions.md
```diff
@@ -0,0 +1,240 @@
...
+    "name": "Demo - Customer A"
+  }
+]
+```
+{% endtab %}
+{% endtabs %}
+
+### StackState UI
+
+#### Check functions
```
mpvvliet, line 224: Actually, logging for check functions is configured on a check level. So, if you have components A and B, each having a stream with a check with check function f, then you can enable logging only on the check of component A or the check of component B. Which will show logging for function f but only if it is invoked from A or from B.

The reasoning behind this is that check functions can be used in many, many places in StackState and enabling debugging on the function itself would generate too much logging.

configure/logging/logging-for-functions.md
```diff
@@ -0,0 +1,240 @@
...
+    - `log.info(variable.toString)`
+    
+2. [Set the logging level](#set-the-logging-level-for-a-function-or-event-handler) for the function.
+
+
+## Set the logging level for a function or event handler
+
+The logging level for individual functions ond event handlers can be set using the StackState CLI. 
+
+1. Find the ID for the function you want to add logging to:
```
mpvvliet, line 26: not "add logging to" but "enable logging for" or "show logging for"

mpvvliet: and indeed, they are not all functions

PR #124, opened 2020-10-29
STAC-10845: Add upgrade note for [4.2]

craffit:


setup/upgrading.md
```diff
@@ -120,6 +120,8 @@ Once StackState has been upgraded and started, verify that the new installation
 * Changes to the `processmanager.conf` and `processmanager/kafka-topics.conf` files must be manually processed if you are using a customised or generated configuration:
   * Added new argument `topology-events` to the section `processes.kafkaToElasticsearch` of file `processmanager.conf`.
   * Added new section `kafka.topics.sts_topology_events` to file `processmanager/kafka-topics.conf`.
+* The etc/stackstate-receiver/application.conf has been changed to have the config in the `stackstate.receiver` namespace. Custom configurations should be adapted.
```
gggina, line 123: This is not clear to me - has configuration been added to the application.conf file or removed? How should custom configurations be adapted?

setup/upgrading.md
```diff
@@ -120,6 +120,8 @@ Once StackState has been upgraded and started, verify that the new installation
...
+  * The apiKey configuration key has been changed to apiKeys
```
gggina, line 124: Where has this key been changed?

setup/upgrading.md
craffit: we 'relocated' a namespace so `stackstate` became `stackstate.receiver` (its this simple string replacement)

setup/upgrading.md
craffit: that was it, on top of the change in namespace

gggina (approved):

PR #193, opened 2020-12-03
Stac 11 servicenow update

gggina:


gggina:
Diagram still requires update. To do together with screenshots for v4.2

PR #196, opened 2020-12-04
Stac 10 review use section

gggina:


gggina:
includes:
- review of use section and configure > telemetry
- async event handler functions
- fix for messy internal links

PR #259, opened 2021-01-26
STAC-11738 How to configure self-signed certificates (4.2.4 release)

rb3ckers:
Part of the StackState 4.2.4 release

This supports a custom trust store that can be used for connecting to applications that use self-signed certificates.

### DIFF
--- a/configure/security/self-signed-certificates.md
+++ b/configure/security/self-signed-certificates.md
@@ -5,13 +5,15 @@ kind: Documentation
 
 # Self-signed certificates
 
-StackState has several points of interaction with external systems, for example event handlers can call out to webhooks in other systems while plugins can retrieve data from external systems like Splunk or Elasticsearch. StackState will not be able to communicate witht these systems when they are secured with TLS using a self-signed certificate or a certificate that is not by default trusted by the JVM.
+StackState has several points of interaction with external systems, for example event handlers can call out to webhooks in other systems while plugins can retrieve data from external systems like Splunk or Elasticsearch. With the default configuration, StackState will not be able to communicate with these systems when they are secured with TLS using a self-signed certificate or a certificate that is not by default trusted by the JVM.
 
 To mitigate this StackState allows configuration of a custom trust store.
 
 ## Creating a custom trust store
 
-You need an installed JVM (the operating system you're on doesn't matter) to be able to create a new trust store. You also need to have the certificate available, if you don't have that [here](#retrieve-certificate-via-the-browser) is a way how to get it.
+To convert an existing TLS certificate file to the format that is needed by StackState, you'll need to use the keytool tool and the cacerts file that both are included in the Java Virtual Machine installation. You can run this on any machine, regardless of the type of operating system.
+
+You also need to have the certificate available, if you don't have that, [retrieve it via the browser](#retrieve-certificate-via-the-browser).
 
 With the JVM installed and the certificate saved as a file `site.cert` you can create a new trust store by taking the JVM's trust store and adding the extra certificate.
 
@@ -20,7 +22,7 @@ With the JVM installed and the certificate saved as a file `site.cert` you can c
    ```bash
    keytool -import -keystore custom_cacerts -alias <a-name-for-the-certificate>  -file site.cert
    ```
-   The alias needs to be a unique alias for the certificate, for example the domain name itself without any dots. Keytool will ask for a password, the JVM default password is `changeit`.
+   The alias needs to be a unique alias for the certificate, for example the domain name itself without any dots. Keytool will ask for a password. The password usually is `changeit`, the JVM default.
 
 The `custom_cacerts` store file will now include the `site.cert` certificate. You can verify that by searching for the alias in the output of
 
@@ -38,7 +40,6 @@ helm upgrade \
   --install \
   --namespace stackstate \
   --values values.yaml \
-  --values authentication.yaml \
   --set-file 'stackstate.java.trustStore'=custom_cacerts \
   --set 'stackstate.java.trustStorePassword'=changeit \
 stackstate \
--- a/configure/security/self-signed-certificates.md
+++ b/configure/security/self-signed-certificates.md
@@ -16,9 +16,17 @@ You need an installed JVM (the operating system you're on doesn't matter) to be 
 With the JVM installed and the certificate saved as a file `site.cert` you can create a new trust store by taking the JVM's trust store and adding the extra certificate.
 
 1. Copy the `cacerts` file from `$JAVA_HOME/lib/security/cacerts` to `custom_cacerts` with `$JAVA_HOME` the location of your Java installation. This environment variable is normally set by the Java installation.
-2. Add the extra certificate(s) with `keytool -import -keystore custom_cacerts -alias <a-name-for-the-certificate>  -file site.cert`. Here you should provide a unique alias for the certificate (for example the domain name itself without any dots). Keytool will ask for a password, the JVM default is `changeit`.
+2. Add the extra certificate(s) with 
+   ```bash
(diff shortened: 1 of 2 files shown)


configure/security/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
+---
+title: Self-signed certificates
+kind: Documentation
+---
+
+# Self-signed certificates
+
+StackState has several points of interaction with external systems, for example event handlers can call out to webhooks in other systems while plugins can retrieve data from external systems like Splunk or Elasticsearch. StackState will not be able to communicate witht these systems when they are secured with TLS using a self-signed certificate or a certificate that is not by default trusted by the JVM.
```
vpartington, line 8: "witht" -> "with"

vpartington: "StackState will not be able to communicate" -> "With the default configuration, StackState will not be able to communicate"

configure/security/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+
+To mitigate this StackState allows configuration of a custom trust store.
+
+## Creating a custom trust store
+
+You need an installed JVM (the operating system you're on doesn't matter) to be able to create a new trust store. You also need to have the certificate available, if you don't have that [here](#retrieve-certificate-via-the-browser) is a way how to get it.
```
vpartington, line 14: no "here" link: "[here](#retrieve-certificate-via-the-browser) is a way how to get it" -> "[retrieve them via the browser](#retrieve-certificate-via-the-browser)"

configure/security/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+
+With the JVM installed and the certificate saved as a file `site.cert` you can create a new trust store by taking the JVM's trust store and adding the extra certificate.
+
+1. Copy the `cacerts` file from `$JAVA_HOME/lib/security/cacerts` to `custom_cacerts` with `$JAVA_HOME` the location of your Java installation. This environment variable is normally set by the Java installation.
```
vpartington, line 18: Give an example command line?

configure/security/self-signed-certificates.md
vpartington: "You need an installed JVM (the operating system you're on doesn't matter)" confused me. How about

"To convert an existing TLS certificate file to the format that is needed by StackState, you'll need to use `keytool` tool that is included in the Java Virtual Machine installation. You can run this on any machine, regardless of the type of operating system"

BTW, what about K8s users that don't know about Java? Shouldn't we offer a Docker-based way to do this?

configure/security/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+2. Add the extra certificate(s) with 
+   ```bash
+   keytool -import -keystore custom_cacerts -alias <a-name-for-the-certificate>  -file site.cert
+   ```
+   The alias needs to be a unique alias for the certificate, for example the domain name itself without any dots. Keytool will ask for a password, the JVM default password is `changeit`.
```
vpartington, line 23: What does it mean that `changeit` is the default? Is that what will be chosen if I press Enter?

configure/security/self-signed-certificates.md
```diff
@@ -0,0 +1,79 @@
...
+The `custom_cacerts` store file will now include the `site.cert` certificate. You can verify that by searching for the alias in the output of
+
+```bash
+keytool -list -keystore custom_cacerts
+```
+
+## Using the custom trust store
+
+### Kubernetes
+For Kubernetes installations the trust store and the password can be specified as values. Since the store is a file it can only be specified from the helm command line. We specify the password value in the same way, but it could also be provided via a `values.yaml` file.
```
vpartington, line 34: Shouldn't we tell users to use the `values.yaml` approach here? `--set-file` and `--set` are messy.

In fact, shouldn't we update the `generate_values.sh` script (I still prefer that over the CLI) to generate this?

vpartington (changes requested):

configure/security/self-signed-certificates.md
rb3ckers: For Linux, WIndows, Mac or all 3?

configure/security/self-signed-certificates.md
rb3ckers: I wouldn't want to polute our values yaml with something only a few customers ever use. I don't know of a way how to put a binary file in a `values.yaml` that is not super messy (i.e. take file, base64 encode and put 50kb of garbage into a text file). So for binary files I always prefer `--set-file` and to not compliate matters more here I give the example completely like this. I'm fine with putting the password into a values file if you prefer that.

configure/security/self-signed-certificates.md
rb3ckers: Serious question though. I don't know what we do in general in the docs.

configure/security/self-signed-certificates.md
rb3ckers: Not sure how far you want to go with that. My experience is that our customers are more likely to have a JVM than to have a docker daemon installed. But I can add that as an alternative.

configure/security/self-signed-certificates.md
rb3ckers: YOu should use the password `changeit`. Rephrased.

configure/security/self-signed-certificates.md
vpartington: I can imagine our current customers are more likely to have a JVM installed. But would that not also be caused by the fact that you need a JVM to run the Linux install for StackState?

But you're right, let's stick to what we need to now and not add even more work to our backlog.

configure/security/self-signed-certificates.md
vpartington: Ah yes. macOS and Linux commands tend to be the same, but I forgot about Windows.

Maybe just give an example for Linux/macOS then?

configure/security/self-signed-certificates.md
vpartington: BTW, is that the password for the keystore (that can be set with `-storepasswd`) or the one for the key itself (`-keypasswd`)? Maybe you add (one of) those two arguments to the command so that the prompt isn't shown and so that it is clear what value to put here and what value to set in the `values.yaml` file?

configure/security/self-signed-certificates.md
vpartington: I see why you'd prefer the `--set-file` option. However, I can imagine users forgetting to add the `--set-file` option when upgrading. The upgrade will appear to succeed, but some of the integrations will start failing later on. Maybe even the alerting is this was used to communicate with an external notification service that was set up with self-signed certs. Unfortunately StS is not very good at checking its own status yet, so it might be a while before they find out. I can imagine some irate tickets coming in when they do find out.

Base64-encoding is ugly indeed, but the `generate_values.sh` script would be able to do that for the user.

configure/security/self-signed-certificates.md
rb3ckers: I now added the linux / mac version (which I think should be the same)

configure/security/self-signed-certificates.md
rb3ckers: Added a docker example as well. I personally prefer that way actually

configure/security/self-signed-certificates.md
rb3ckers: I didn't even know about the `keypasswd`. This is the default password on the store itself. `keytool` always asks this automatically.

configure/security/self-signed-certificates.md
rb3ckers: Let's discuss this offline like you suggested.

rb3ckers: Discussed and we've decided to stick with the `set-files`

PR #261, opened 2021-01-27
STAC-11770 Add documentation on Elasticsearch privileged containers

rb3ckers:


setup/installation/kubernetes_install/required_permissions.md
```diff
@@ -0,0 +1,91 @@
...
+
+If it is not at least 262144 it needs to be increased in a different way.
+
+## Increasing Linux system settings for Elasticsearch
+
+Depending on what your Kubernetes administrators prefer they could either set the `vm.max_map_count` to a higher default on all nodes by changing the default node configuration (for example via init scripts)  or by having a daemonset do this right after node startup. The former is very dependent on your Kuberentes cluster setup so there are no general solutions there.
+
+However here is an example that can be used as a starting point for a daemonset to change the setting:
+
+```yaml
```
rb3ckers, line 28: I think it is a good idea to have this as an example here. If you agree I'll run a test for this as well (not yet verified this works properly, but it should)

vpartington: Wow, that is pretty elaborate. 

And probably a good idea because if we don't provide this, we'll either get asked how to do this or people don't do it which will lead to ElasticSearch issues later on.

Or does ElasticSearch check that the values have been set correctly at startup? If that is the case, you might want to mention that explicitly, including the error message so that administrators will be able to quickly find this section if they forget to set it or the config breaks in some other way.

vpartington (changes requested):
Dependening on the answer to my question, you might want to add a paragraph that explains what happens if the administrators disables the sysctlInitContainer but does not perform the other actions.

setup/installation/kubernetes_install/required_permissions.md
rb3ckers: I thought I mentioned that already, but if not will add it and add the error message ES gives. Because it does check on starup and will just fail to startup with a very clear error message saying something along the lines `vm.max_map_count is abc but should be greater than or equal to xyz`.

rb3ckers: Added example error

setup/installation/kubernetes_install/required_permissions.md
```diff
@@ -0,0 +1,95 @@
+# Required permissions
+All of StackState's own components can run without any extra permissions. However StackState uses Elasticsearch which has some extra requirements on the nodes it runs on. To make sure those requirements are satisfied it uses an init container that runs in privileged mode and as the root user to change the `vm.max_map_count` Linux system setting on the nodes it is started on. This init container is enabled by default, because the setting usually is lower than required and would not allow Elasticsearch to start.
```
rb3ckers, line 2: The sentence from "To make sure" is a bit weird and hard to understand I just noticed

PR #265, opened 2021-02-05
Update URL builder API with a bunch of new methods (4.3)

Q69K:


develop/reference/scripting/script-apis/ui.md
```diff
@@ -30,25 +30,74 @@ No arguments.
...
+
+```groovy
+View.getAll().then { views ->
+    Component.withId(component).get().then { component ->
+        UI.createUrl().view(views[0]).at('-15m').withComponent(component).url()
+    }
+}
+```
+
+Create a URL to topology perspective with all components from production environment that in critical state with full root cause tree.
```
gggina, line 95: "Create a URL to the topology perspective explore mode with filters in place to show all components from the production environment that are in critical state and showing the full root cause tree."

Q69K: Sounds good, thank you!)
```suggestion
Create a URL to the topology perspective explore mode with filters in place to show all components from the production environment that are in critical state and showing the full root cause tree
```

gggina (approved):

PR #327, opened 2021-03-18
Stac 11 kubernetes sp new

gggina:


stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+## Overview
+
+The Kubernetes integration is used to create a near real-time synchronization of topology and associated internal services from a Kubernetes cluster to StackState. This StackPack allows monitoring of the following:
+
+* Workloads
+* Nodes, pods, containers and services
+
+![Data flow](/.gitbook/assets/stackpack_kubernetes_draft2.svg)
+
+The Kubernetes integration collects topology data for nodes, pods, containers and services in a Kubernetes cluster as well as metrics and events.
```
rb3ckers, line 17: Also topology data for workklads, configmaps, secrets and volumes.

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+
+- StackState Agent V2 is deployed **on each node** in the Kubernetes cluster:
```
rb3ckers, line 19: After `**on each node**` I would add that this is a `DaemonSet`.

rb3ckers: Maybe like this: "StackState Agent V2 is deployed as a DaemonSet with one instance **on each node** in the Kubernetes cluster:"

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+    * Host information is retrieved from the Kubernetes API
+    * Container information is collected from the Docker daemon
+    * Metrics are retrieved from kubelet running on the node and also from kube-state-metrics if this is deployed on the same node
+    * Events ???
+- StackState Cluster Agent is deployed **on one node** in the Kubernetes cluster:
```
rb3ckers, line 24: Here I would omit the `**on one node**` remark. The only place where nodes are really relevant in k8s is for DaemonSets, here it is mostly confusing I think. 

Maybe instead: "StackState Cluster Agent is deployed with a Deployment with one instance for the entire cluster."

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+    - Relevant [metrics data](#metrics) is mapped to associated components and relations in StackState. All retrieved metrics data is stored and accessible within StackState.
+    - [Events](#events) are available in the StackState Events Perspective and listed in the details pane of the StackState UI.
+
+## Setup
+
+### Prerequisites
+
+The following prerequisites are required to install the Kubernetes StackPack and deploy the StackState Agent and Cluster Agent:
+
+* A Kubernetes Cluster must be up and running, and you must be able to connect to it from StackState.
```
rb3ckers, line 40: `and you must be able to connect to it from StackState.` : StackState is not yet installed, so this is not the case. However the user must be able to connect with it to run the helm install command. Actually there is one more additional requirement and that is that the user needs quite significant permissions. So I think we should make that more explicit:

"A Kubernetes Cluster must be up and running, and you must have permissions to create privileged pods and create ClusterRoles and ClusterRoleBindings."

This may actually raise some questions I guess, so we may have to explain these permissions as well:
* ClusterRole and ClusterRoleBinding are needed such that the agents are granted permissions to access the Kubernetes API
* The Agent V2 gathers information on network connections and host information which both require it to run in a privileged pod

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+    * `stackstate.url`
+    * It is also recommended to provide a `stackstate.cluster.authToken`. This is an optional variable, however, if one is not provided a new, random value will be generated each time a helm upgrade is performed. This could leave some pods in the cluster with an incorrect configuration.
+
+```
+helm install \
+--set-string 'stackstate.apiKey'='<your-api-key>' \
+--set-string 'stackstate.cluster.name'='<your-cluster-name>' \
+--set-string 'stackstate.cluster.authToken'='<your-cluster-token>' \
+--set-string 'stackstate.url'='<your-stackstate-url>' \
+stackstate/cluster-agent
```
rb3ckers, line 78: Just realized that in other places in the docs we tell people to create a `values.yaml` file that contains these instead.  Maybe we should do that here as well? It would look like this:

```yaml
stackstate:
  apiKey: <your-api-key>
  cluster:
    name: <your-cluster-name>
    authToken: <your-cluster-token>
  url: <your-stackstate-url>     
```

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+
+1. If you do not already have it, you will need to add the StackState helm repository to the local helm client:
+    ```
+    helm repo add stackstate https://helm.stackstate.io
+    helm repo update
+    ```
+2. Deploy the StackState Agent, Cluster Agent and kube-state-metrics with the helm install command below. The following variables should be provided:
+    * `stackstate.apiKey`
+    * `stackstate.cluster.name`
+    * `stackstate.url`
```
rb3ckers, line 69: These 3 can all be found when installing the StackPack. Should we for the docs refer to that?

stackpacks/integrations/kubernetes_new.md
rb3ckers: Oh but like you have it now may be nicer, because the StackPack already provides an example command line that can be copy/pasted. We need to tweak that anyway to show this command instead, but in that case we can fill in ALL of these for the user and they just need to copy/paste.

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+--set-string 'stackstate.cluster.authToken'='<your-cluster-token>' \
+--set-string 'stackstate.url'='<your-stackstate-url>' \
+stackstate/cluster-agent
+```
+
+Full details of the available values can be found in the [Cluster Agent Helm Chart documentation \(github.com\)](https://github.com/StackVista/helm-charts/tree/master/stable/cluster-agent).
+
+### Status
+
+To check the status of the Kubernetes integration, check that the StackState Cluster Agent (`cluster-agent`) pod and all of the StackState Agent (`cluster-agent-agent`) pods have status ready.
```
rb3ckers, line 85: In the example we assume they installed in the `monitoring` namespace. So if people copy this without thinking (which they do) it may not work. So we probably should add the namespace also to the helm command as well and add a remark that it can be changed of course.

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+| **Others** | All other events |
+
+#### Metrics
+
+The Kubernetes integration makes all metrics from the Kubernetes cluster available in StackState. Relevant metrics are automatically mapped to the associated components. 
+
+All retrieved metrics can be browsed or added to a component as a telemetry stream. Select the data source **StackState Metrics** and type `kubernetes` in the **Select** box to get a full list of all available metrics. 
+
+![Add a Kubernetes metrics stream to a component](/.gitbook/assets/v43_add_k8s_data_stream.png)
+
```
rb3ckers, line 125: Would be nice to have a picture that shows some data

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+#### Traces
+
+The Kubernetes integration does not retrieve any traces data.
+
+#### Tags
+
+All tags defined in Kubernetes will be retrieved and added to the associated components and relations in StackState. 
+
+### REST API endpoints
```
rb3ckers, line 140: There are only 2 endpoints in the list. All the others are "resource types". Is there a specific reason for including this at all? As a user of the StackPack this is mostly uninteresting and I'd be more interested in the component actions and views I get

stackpacks/integrations/kubernetes_new.md
```diff
@@ -0,0 +1,285 @@
...
+* Kubernetes - Infrastructure - <instance_name>
+* Kubernetes - Namespaces - <instance_name>
+* Kubernetes - Workload Controllers - <instance_name>
+
+
+### Open source
+
+The code for the StackState Agent Kubernetes check is open source and available on GitHub at:
+
+- [https://github.com/StackVista/stackstate-agent-integrations/tree/master/kubernetes](https://github.com/StackVista/stackstate-agent-integrations/tree/master/kubernetes)
```
rb3ckers, line 197: These `check`s are only a very tiny piece of the code. Most of it actually lives here inside the agent itself: https://github.com/stackvista/stackstate-agent

Or more specifically here: https://github.com/StackVista/stackstate-agent/tree/master/pkg/collector/corechecks/cluster

stackpacks/integrations/kubernetes_new.md
gggina: This became a long list. I have added "Configmaps, secrets and volumes" above the diagram and removed all specific info from this line.

stackpacks/integrations/kubernetes_new.md
gggina: done

stackpacks/integrations/kubernetes_new.md
gggina: "StackState Cluster Agent is deployed with a Deployment. There is one instance for the entire Kubernetes cluster"

stackpacks/integrations/kubernetes_new.md
gggina: I have changed to request they use the command provided and added a note explaining the `stackstate.cluster.authToken`

gggina: Note that this does mean the stackpack UI docs must be updated at the same time as the docs page goes live.

stackpacks/integrations/kubernetes_new.md
gggina: I used thename `monitoring`
```
helm install \
--namespace monitoring \
--set-string 'stackstate.apiKey'='<your-api-key>' \
--set-string 'stackstate.cluster.name'='<your-cluster-name>' \
--set-string 'stackstate.cluster.authToken'='<your-cluster-token>' \
--set-string 'stackstate.url'='<your-stackstate-url>' \
stackstate/cluster-agent
```

stackpacks/integrations/kubernetes_new.md
gggina: there is a data chart now

stackpacks/integrations/kubernetes_new.md
gggina: Added the two extra links

stackpacks/integrations/kubernetes_new.md
rb3ckers: :+1:

stackpacks/integrations/kubernetes_new.md
rb3ckers: :+1:

gggina:
will switch PR to master branch

PR #360, opened 2021-04-12
Adds note box to 4.2 docs that this is an old version

gggina:


gggina:
making new branch without gitbook changes to assets

PR #421, opened 2021-05-05
STAC-12894: document how to create a check function

fvlankvelt:


configure/telemetry/checks_and_streams.md
```diff
@@ -46,11 +46,85 @@ A baseline can be attached to a metric stream. The baseline consists of an avera
...
+```text
+if (metrics[-1].point >= deviatingValue) return DEVIATING;
+```
+{% endtab %}
+{% endtabs %}
+
+## Check Result
+
+When a check function runs, it returns a check result.  This result can be either one of
+* **health state** a `HealthStateValue`, the new health state of the component.  One of `CLEAR`, `DEVIATING`, `CRITICAL`, `FLAPPING`, `DISABLED` or `UNKNOWN`
```
konstantintatarnikov, line 85: Do we really interpret FLAPPING state anywhere in stackstate? (I know that we do have it in the enum).

configure/telemetry/checks_and_streams.md
```diff
@@ -46,11 +46,85 @@ A baseline can be attached to a metric stream. The baseline consists of an avera
...
+
+| Key | Type | Description |
+| :--- | :--- | :--- |
+|`runState` |`RunStateValue`|see above|
+|`healthState` |`HealthStateValue`|see above|
+|`healthAcknowledgement` |`CheckAcknowledgementResult`|see above|
+|`expiration` |`CheckStateExpiration`|see above|
+|`detailedMessage` |`String`|detailed explanation of the reason behind the health status|
+|`shortMessage` |`String`|a short description|
+|`causingEvents` |`List<EventRef>` or `List<Map<String, Object>>`|the events that triggered the health state change|
```
konstantintatarnikov, line 101: I am wondering if we should mention, that causingEvent is just a list of maps...
|`causingEvents` |`List<Map<String, Object>>

gggina: Can it only be a `List<Map<String, Object>>` or also `List<EventRef>` ?

configure/telemetry/checks_and_streams.md
```diff
@@ -46,11 +46,85 @@ A baseline can be attached to a metric stream. The baseline consists of an avera
...
+```
+{% endtab %}
+{% endtabs %}
+
+## Check Result
+
+When a check function runs, it returns a check result.  This result can be either one of
+* **health state** a `HealthStateValue`, the new health state of the component.  One of `CLEAR`, `DEVIATING`, `CRITICAL`, `FLAPPING`, `DISABLED` or `UNKNOWN`
+* **run state** a `RunStateValue`, the new run state of the component.  One of `UNKNOWN`, `STARTING`, `RUNNING`, `STOPPING` `STOPPED`, `DEPLOYED` or `DEPLOYING`
+* **acknowledgement**, a `CheckAcknowledgementResult`
```
gggina, line 87: What is the acknowledgement?

configure/telemetry/checks_and_streams.md
```diff
@@ -46,11 +46,85 @@ A baseline can be attached to a metric stream. The baseline consists of an avera
...
+
+Check functions make it possible to configure a health check that triggers the DEVIATING health status.  They do this by processing metric data, logs or events. You can write your own check function as a groovy script. To add a custom event handler function:
+
+1. Go to **Settings** &gt; **Functions** &gt; **Check Functions**.
+2. Click **ADD CHECK FUNCTION**.
+3. Enter the required settings:
+  * **Name** - A name to identify the event handler function.
+  * **Description** - Optional. A description of the event handler function.
+  * **User parameters** - parameters that must be entered by the user when a check is added to a component. For details, see the section on [parameters](checks_and_streams.md#parameters) below.
+  * **Script** - The script run by the function.
```
gggina, line 59: Doesn't cover Return: Health state/ run state

configure/telemetry/checks_and_streams.md
```diff
@@ -46,11 +46,85 @@ A baseline can be attached to a metric stream. The baseline consists of an avera
...
+{% endtab %}
+{% endtabs %}
+
+## Check Result
+
+When a check function runs, it returns a check result.  This result can be either one of
+* **health state** a `HealthStateValue`, the new health state of the component.  One of `CLEAR`, `DEVIATING`, `CRITICAL`, `FLAPPING`, `DISABLED` or `UNKNOWN`
+* **run state** a `RunStateValue`, the new run state of the component.  One of `UNKNOWN`, `STARTING`, `RUNNING`, `STOPPING` `STOPPED`, `DEPLOYED` or `DEPLOYING`
+* **acknowledgement**, a `CheckAcknowledgementResult`
+* **expiration**, a `CheckStateExpiration` that specifies how long the health state remains valid and what it should expire to
```
gggina, line 88: only health state or also run state?

configure/telemetry/checks_and_streams.md
fvlankvelt: could very well be that we indeed to not.  There are some TODO comments.

configure/telemetry/checks_and_streams.md
fvlankvelt: if there's no easy way to get an `EventRef` from an event, then it is indeed just confusing.

configure/telemetry/checks_and_streams.md
fvlankvelt: I'm not sure tbh.  @konstantintatarnikov ?

configure/telemetry/checks_and_streams.md
fvlankvelt: only health state - I also can imagine that run states (e.g. RUNNING or STOPPED) simply stay valid until there is an event.  Whereas health requires more active monitoring.

fvlankvelt:
from @lowdewijk 
* The CheckResult return type is unclear to me. It needs examples. Can we provide an example with all values set? I think this description in terms of the types is very hard to follow.
* I would also for the types not say “see above”.  And then for causing events there is effectively a. “see below”. Simply add it to the description field.
*This part of “In order to guarantee that data can be serialized and no objects can leak from the groovy sandbox, only these value types are supported:” I would simply not state that. It should simply be clear how to construct the return type.
* It also not necessary for the user to understand that the return type is a CheckResult . It is simply the return type of the check.
* I would drop the healthAcknowledgement from the docs. Nobody uses this function, thus I think we can completely drop it and I do not want people to start asking us questions about it.
detailedMessage is actually interpreted as markdown.
* For paramters we also have log streams and StackState event stream.
* For the identifier field we can link that to a place where the identifiers are explained.

develop/developer-guides/custom-check-function.md
```diff
@@ -0,0 +1,72 @@
...
+
+User parameters can be any of the types listed below. **Required** parameters must always be specified by the user. **Multiple** parameters will be a list containing the specified user parmameter type.
+
+| User parameter type | Description |
+|:---|:---|
+| Primitive types | Can be a  `Float`, `Integer`, `String` or `Boolean`. |
+| State | |
+| Run state | |
+| Metric stream | A `List` with a number of `MetricTelemetryPoint` metrics. Each `MetricTelemetryPoint` includes a `point` (the metric value as a double) and a `timestamp` in epoch milliseconds. | 
+| Baseline metric stream | |
```
fvlankvelt, line 37: these are deprecated, probably good to note here

develop/developer-guides/custom-check-function.md
```diff
@@ -0,0 +1,72 @@
...
+{% endtab %}
+{% endtabs %}
+
+## Result
+
+Whenever a check function runs, it returns a result. This can be a **health state** and/or a **run state** or a custom map containing a collection of data formatted as descibed below. 
+
+* **Health state** - A `HealthStateValue`. This will be the new health state of the component (`CLEAR`, `DEVIATING`, `CRITICAL`, `DISABLED` or `UNKNOWN`). A `CheckStateExpiration` can also be returned to specify how long the health state should remain valid and what it should change to after expiration.
+* **Run state** - A `RunStateValue`. This will be the new run state of the component (`UNKNOWN`, `STARTING`, `RUNNING`, `STOPPING` `STOPPED`, `DEPLOYED` or `DEPLOYING`). |
+* **Custom map** - A custom map can contain a health state and/or run state as described above as well as:
```
fvlankvelt, line 59: I think the `expiration` has gone missing here?

develop/developer-guides/custom-check-function.md
```diff
@@ -0,0 +1,72 @@
...
+  - `detailedMessage` - Markdown formatted explanation of the reason behind a returned health state. `String`.
+  - `shortMessage` - A short description of the state change. `String`.
+  - `causingEvents` - The events that triggered the health state change. These are used in [anomaly check functions](/develop/developer-guides/anomaly-check-functions.md) to link changes to anomaly events. Provided as a map with the keys `title` (`String`), `eventId` (`String`), `eventTimestamp` (`Long`) and `eventType` (`String`).
+  - `data` - Arbitrary additional data. `Map<String, Object>`.
+
```
fvlankvelt, line 64: a complete example:
```
[
    runState: STOPPING,
    healthState: DEVIATING,
    shortMessage: "Something bad happened",
    detailedMessage: "Something **REALLY** bad happened.",
    expiration: [
        duration: 60000,
        expireTo: UNKNOWN
    ],
    causingEvents: [
         [
                 title: event.getName(),
                 eventId: event.getIdentifier(),
                 eventTimestamp: event.getEventTimestamp(),
                 eventType: event.getType()
         ]
    ],
    data: [
        "foo": "bar"
    ]
]
```

develop/developer-guides/custom-check-function.md
```diff
@@ -0,0 +1,72 @@
...
+## User parameters
+
+Check functions can work with a number of user parameters. These parameters can be accessed by the function script, and their values are specified by the user when the check function is used in a health check.
+
+User parameters can be any of the types listed below. **Required** parameters must always be specified by the user. **Multiple** parameters will be a list containing the specified user parmameter type.
+
+| User parameter type | Description |
+|:---|:---|
+| Primitive types | Can be a  `Float`, `Integer`, `String` or `Boolean`. |
+| State | |
```
fvlankvelt, line 34: this concerns health states.  (should probably be renamed in the UI)

develop/developer-guides/custom-check-function.md
gggina: I added it to the health state description as it was only relevant to health state

develop/developer-guides/custom-check-function.md
gggina: Will do. Can you provide an explanation for each of the other types?

develop/developer-guides/custom-check-function.md
gggina: Can you provide an explanation for each user parameter type?

PR #438, opened 2021-05-17
[4.4] health synchronization

gggina:
Not to be merged just now !


configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
+# Health synchronization
+
+## Overview
+
+One of the ways to get health checks into StackState topology elements is by synchronizing the data from an external system that has already calculated them with it's on data and rules. StackState allows you to automatically synchronize the health state from your external system into StackState and tor relate it to your topology elements ingested via [Topology synchronization](../topology/topology_synchronization.md)
```
craffit, line 5: i would call external system => existing monitoring solution. this will resonate more with the readers i think
type: tor

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+A health synchronization is completely defined and controlled by the data you [send to our Receiver API](send-health-data.md) meaning that there is no extra configuration to apply into StackState itsels, to this we refer as a `zero settings` synchronization.
```
craffit, line 6: itsels=itself. I would not use the word `zero settings` here (this is more for us internal). You could phrase this like: health information can be synchronized by sending data to the receiver api.
(and leave it at that)

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Synchronization pipeline
+The health synchronization framework works by first ingesting the health data sent to the Receiver API into an `External health model` and in a second step trying to find topology elements that related to the just ingested health checks, the relation if formed based on the topology [identifiers](../topology/sync.md#id-extraction) obtained during the topology synchronization and the [topologyElementIdentifier](send-health-data.md#health-json) part of the health payload. StackState keeps track of all the topology element changes as well as in the `External health model` to update the information on stackState's topology.
```
craffit, line 9: i don't think we need to explain the first step in the `External health model`. (Maybe we can ditch `External health model altogether). All the user needs to know is what you write in the middle, the topologyelementIdentifier is matched with the idnetifier form topology synchornization

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+
+The concepts in the synchronization pipeline are explained here:
+### Stream and SubStream
```
craffit, line 12: nice

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Repeat Interval
+As previously mentioned StackState health synchronization process the checks data in a snapshots per SubStream, the repeat interval is the commitment from the external system to send complete snapshots over and over in order to keep the data up to date on StackState. This is helpful for StackState in order to inform the user how uo to date is the health synchronization running.
+
+### Expire Interval
+The expire interval is the way to configure SubStreams the health synchronization to delete data thats is not sent by the external system anymore, it's helpful when the agent of the specific SubStream could potentially be decommissioned and StackState would not here from it gaian leaving it's previosuly synchronized data hanging permanently.
+
+{% hint style="info" %}
+The health expire feature is still under development, so currently the `expiry_interval_s` value is not used at all.
```
craffit, line 22: super

configure/health/send-health-data.md
```diff
@@ -0,0 +1,144 @@
...
+  "internalHostname": "localdocker.test",
+  "metrics": [],
+  "service_checks": [],
+  "topologies": [],
+  "health":[
+  [
+    {
+      "start_snapshot": {
+        "repeat_interval_s": 50
+        //"expiry_interval_s": 200 Optional
```
craffit, line 112: i doubt this curl command will work with the comments inthere

use/health-state-and-event-notifications/health-state-in-stackstate.md
```diff
@@ -0,0 +1,21 @@
...
+
+* Green - `CLEAR` - There is nothing to worry about.
+* Orange - `DEVIATING` - Something may require your attention.
+* Red - `CRITICAL` - Attention is needed right now, because something is broken.
+* Gray - `UNKNOWN` - The view does not have a view health state.
+
+The health checks that a topology element can contain can be obtained in two different ways:
+
+* Calculated by StackState based on telemetry or event streams defined for the topology element. See [Add a health check](../health-state-and-event-notifications/add-a-health-check.md). This path is useful when the telemetry/events data is alredy sent into StackState which opens the posibiliti to use [AAD](../health-state-and-event-notifications/anomaly-health-checks.md)  as well on top of it.
+* Calculated by an external system based on its own rules and synchronized to StackState. See [Set up a health synchronization](../../configure/health/health-synchronization.md). This path is useful when the health checks are already defined and calculated externally or sending the telemetry/events and translating the check rules into StackState is not viable.
```
craffit, line 15: i would again drop the word external monitoring system inhere somehwere

configure/health/debug-health-sync.md
```diff
@@ -0,0 +1,110 @@
...
+{% endtab %}
+{% endtabs %}
+
+{% hint style="info" %}
+The substream status provides useful information to verify that the check states send from the external system into StackState could be related and linked to existing topology elements. In the example we show a `Check states with identifier which has no matching topology element` a case where a check state `checkStateId2` could not be related to a topology element identified as `server-2`.
+{% endhint %}
+
+
+### Error messages
+* **StreamMissingSubStream** - Raised when the health synchronization receives messages with a previous start snapshot in place.
```
craffit, line 95: These are actually never printed, maybe we should do that..

configure/health/debug-health-sync.md
```diff
@@ -0,0 +1,110 @@
+# Debug health synchronization
+
+## Overview
+StackState health synchronization provides a series of error messages via the [StackState CLI](/setup/installation/cli-install.md)] in order to debug and fix issues that might be preventing health data to be correctly ingested and displayed on StackState.
```
craffit, line 4: This page is ok like this, it might be easier for the user to digest if we write some problem-centric header like:
- An check state i am sending does not end up on the component, what to do?
- It takes a long time for a check state to change state in stackstate, where to look?

These are just suggestions, we can also keep it like this and treat it more as reference

configure/health/health-synchronization.md
aacevedoosorio: done

configure/health/health-synchronization.md
aacevedoosorio: done

configure/health/send-health-data.md
aacevedoosorio: removed the comments so the curl will work

use/health-state-and-event-notifications/health-state-in-stackstate.md
aacevedoosorio: changed it for existing monitoring solution

configure/health/debug-health-sync.md
aacevedoosorio: I'll start printing them in the story when I implement expiry. And in the scope of that story potentially update this part of the docs

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+
+The concepts in the synchronization pipeline are explained here:
+### Stream and SubStream
+The stream uniquely indentifies the health synchronization and defines the boundaries of which check states should be processed together in strict order, within the stream we have the SubStream in which the snapshot lifecyles are applied. The SubStream is the way StackState offers to model when the agents to report health for a single stream are distributed in different locations and which data is semi independent form each other as well as their own snapshots but in the end they all contribute to the check states for a stream. It's possible to model a health synchronization without defining a SubStream, for example when a single agent is responsible of reporting the check states for a stream, in that case the `sub_stream_id` can be omitted and StackState will assume all the external checks belong to a single default SubStream.
```
gggina, line 16: Not clear on:
- what is meant here by "agent" ?
- "The SubStream is the way StackState offers to model when the agents to report health for a single stream are distributed in different locations and which data is semi independent form each other as well as their own snapshots but in the end they all contribute to the check states for a stream."
- does (can) a stream contain multiple sub-streams?

configure/health/health-synchronization.md
```diff
@@ -0,0 +1,29 @@
...
+As previously mentioned StackState health synchronization process the checks data in a snapshots per SubStream, the repeat interval is the commitment from the external system to send complete snapshots over and over in order to keep the data up to date on StackState. This is helpful for StackState in order to inform the user how uo to date is the health synchronization running.
+
+### Expire Interval
+The expire interval is the way to configure SubStreams the health synchronization to delete data thats is not sent by the external system anymore, it's helpful when the agent of the specific SubStream could potentially be decommissioned and StackState would not here from it gaian leaving it's previosuly synchronized data hanging permanently.
+
+{% hint style="info" %}
+The health expire feature is still under development, so currently the `expiry_interval_s` value is not used at all.
+{% endhint %}
+
+### Check State
```
gggina, line 28: Is this correct: the synchronization provides a health state (+ message) that applies to a single component and is the equivalent of configuring + running a health check on an external system

configure/health/debug-health-sync.md
aacevedoosorio: I think the `Solutions to common issues` aligns with what Bram suggested with the problem-centric approach ...
The `Error messages` have their own section already there

PR #472, opened 2021-06-04
Fix typo found in zendesk#806

hierynomus:


hierynomus:
Should be ported to 4.3 (and maybe 4.2?)

gggina (approved):

PR #504, opened 2021-06-25
remove all docs references to baselines

gggina:


configure/topology/propagation.md
```diff
@@ -1,44 +0,0 @@
----
-description: Propagate health states from dependencies to dependents
----
```
konstantintatarnikov, line 3: Why Propagation function doc has been removed?

develop/developer-guides/functions.md
```diff
@@ -1,41 +0,0 @@
-# Overview
-
-Functions in StackState are predefined scripts that transform input into an output. They are called by StackState on-demand. For example, when a component changes state, new telemetry flows in ora user triggers an action. Advanced users can develop their own functions to customize StackState. These functions can then be exported and packaged with a custom StackPack. Read more about it in [how to create StackPacks](../../stackpacks/about-stackpacks.md).
```
konstantintatarnikov, line 3: hmm , this is just general information about functions - not about baselines, should we keep it?

configure/topology/propagation.md
gggina: this was an old, unused page.
it is now /develop/developer-guides/custom-functions/propagation-functions.md

develop/developer-guides/functions.md
gggina: this is an archived page. it was not used

konstantintatarnikov (approved):

PR #524, opened 2021-07-01
STAC-13269 Golden signals

julianosk:


use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+
+The `Error percentage` check function monitors two streams - one reporting errors and one reporting a total. A DEVIATING health state is returned when the percentage of errors/total crosses the specified `DeviatingThresholdPercentage` and a CRITICAL health state is returned when the error percentage crosses the specified `CriticalThresholdPercentage`.
+
+If your SLO defines that a service can have a maximum of 5% of requests failing, you can create a check using the `Error percentage` function and set the `Critical Threshold Percentage` to `5.0`:
+
+![Error percentage check](../../images/telemetry/http-error-check.png)
+
+### Check function: Total response time check
+
+
```
gggina, line 82: I couldn't find a description of this check function.

julianosk: Changed to `Greater than or equal` check function

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+## Checks
+
+To help you meet your SLA (Service Level Agreement) you can create [checks](/use/health-state-and-event-notifications/health-state-in-stackstate.md#health-checks) in StackState. Examples of using a check function to monitor error percentage and response time are given below.
+
+When selecting a metric stream for a health check, you will have some options to configure its behavior:
+
+- Windowing method: [more details in this page](../health-state-and-event-notifications/add-a-health-check.md#windowing-method).
+- Aggregation: [see here a list of all possible aggregation methods](../../develop/reference/scripting/script-apis/telemetry.md#aggregation-methods).
+- Time window (or window size): By default the time window is 300000 milliseconds (or 5 minutes). The greater the time window is, less frequent your health state will be updated. Though if you configure it too small, you might have too many state changes for the component.
```
rb3ckers, line 70: In case of a sliding window this is not true. Then the check will update on each new value received. What is affected by this time window is the number of false positive / false negative alerts you will get. The shorter the time window the more sensitive it is going to be to a sudden spike in errors (which is not what you'd want for alerts on an SLO). The longer you make it the less sensitive you'll be.

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
+# Golden signals
+
+## Overview
+
+When monitoring distributed systems with a defined SLO (Service Level Objective), StackState can alert you if one of your SLIs (Service Level Indicators) falls bellow a certain threshold. The [four golden signals \(sre.google\)](https://sre.google/sre-book/monitoring-distributed-systems/#xref_monitoring_golden-signals) described on this page can be used to build a check in StackState that responds to fluctuations in service level.
```
rb3ckers, line 5: I would add a sentence to make clear that these checks don't check your SLO directly. SLO's are usually defined on SLI's over an entire month (not minutes/hours/days). Using the short time windows as described in this page (I wouldn't advice using very long ones for alerting) will only help a user in catching potential SLO violations early on and prevent them. With the risk of false positives of course.

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Check function: Greater than or equal
+
+The `Greater than or equal` check function can alert you when one of your telemetry streams is above a certain threshold. A DEVIATING health state is returned when `DeviatingThreshold` is crossed and a CRITICAL health state is returned when the metric stream value is above the specified `CriticalThreshold`.
+
+Use this function to make sure you meet your SLO for maximum response time, for example:
+
+![Response time check](../../images/telemetry/http-resp-time-check.png)
+
+## Limitations
```
rb3ckers, line 88: Next to the limitations I would add a small section explaining that StackState agent will gather these metrics by default for HTTP traffic but that it can be disabled (maybe refer to the agent page for that)

rb3ckers: Also users can load the trace agent in their application and that will usually give them response times metrics again but now from the trace agent (hopefully that is documented somewhere else already :crossed_fingers: )

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+
+![Error percentage check](../../images/telemetry/http-error-check.png)
+
+### Check function: Greater than or equal
+
+The `Greater than or equal` check function can alert you when one of your telemetry streams is above a certain threshold. A DEVIATING health state is returned when `DeviatingThreshold` is crossed and a CRITICAL health state is returned when the metric stream value is above the specified `CriticalThreshold`.
+
+Use this function to make sure you meet your SLO for maximum response time, for example:
+
+![Response time check](../../images/telemetry/http-resp-time-check.png)
```
rb3ckers, line 86: I  see that we have the `Max` aggregation on the stream (which will be used for the check automatically as well). This is not ideal because now the check will go off if there is at least one value over the threshold (and even worse I think, it will remain critical for 5 minutes, or whatever the time window is, after going back to normal).

Had we used the `Min` aggregation in combination with the `greater than` check it would work much nicer. It would need at least a sustained period of slow responses  and would go back to clear as soon as the error rate drops again. This is usually preferred because it gives you a lot less false positives (and if you keep the time window short enough you won't get many false negatives). This is actually our poor-mans implementation of Prometheus' `for: 5 minutes` configuration option for checks.

julianosk: Should we update the stackpack then?

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+
+- Windowing method: [more details in this page](../health-state-and-event-notifications/add-a-health-check.md#windowing-method).
+- Aggregation: [see here a list of all possible aggregation methods](../../develop/reference/scripting/script-apis/telemetry.md#aggregation-methods).
+- Time window (or window size): By default the time window is 300000 milliseconds (or 5 minutes). The greater the time window is, less frequent your health state will be updated. Though if you configure it too small, you might have too many state changes for the component.
+
+### Check function: Error percentage
+
+The `Error percentage` check function monitors two streams - one reporting errors and one reporting a total. A DEVIATING health state is returned when the percentage of errors/total crosses the specified `DeviatingThresholdPercentage` and a CRITICAL health state is returned when the error percentage crosses the specified `CriticalThresholdPercentage`.
+
+If your SLO defines that a service can have a maximum of 5% of requests failing, you can create a check using the `Error percentage` function and set the `Critical Threshold Percentage` to `5.0`:
```
rb3ckers, line 78: The relation is not as one-on-one as suggested here. Not sure how to fix it though without adding too much confusion. I'll try to explain but am afraid we'll just have to leave it as-is or maybe tweak it a little bit.

Your SLO usually is not just a percentage but it also defines a period over which it is measured. Usually this is a month, a quarter or a year. So this check as described here will typically alert way before you are in danger of missing the SLO targets. So in general you may want to put the threshold higher than the value in your SLO.

Additionally an SLO and SLA usually define not a maximum error percentage but an availability percentage. For example 99.9% availability per month (note that this has a slightly different meaning than  99.9% availability per year). That gives roughly 7 hours downtime per month but you can also translate it to 99.9% of the requests should be resolved with a proper response (i.e. not a HTTP 5xx). So over a month only 0.1% of the requests may fail with 5xx errors. Putting that on the check as the threshold for 5 minutes would be a bit too aggressive (one person retrying his failing request may wake your on-call). The SRE workbook has a whole explanation on how to solve this problem with error budgets and their burn-rates giving you early alerts when needed but not too many while also reacting quickly to the errors being resolved. If you have some time I recommend reading this: https://sre.google/workbook/alerting-on-slos/. Problem is we don't have that flexibility in StackState right now.

use/metrics-events/golden_signals.md
```diff
@@ -0,0 +1,90 @@
...
+
+![Error percentage check](../../images/telemetry/http-error-check.png)
```
rb3ckers, line 80: See my comment on the Response time check. Something similar applies here even though we have the average on the stream. Having The `max` on the # request and the min on the # errors would get us close to a similar behavior where we will only alert if the problematic situation sustains for the time window or longer (again mimicing `for: 5 minutes` from Prometheus). 

The problem is of course that this messes with the visualisation of the streams in the normal charts and makes them hard to compare (the current aggergations are better for that). 

So for these cases I guess we want to keep the current aggergations in the stackpack and therefore compromise (even more) on the checks. We can check with Mark Bakker though: "Where do you want to compromise: streams in UI or checks?"

use/metrics-events/golden_signals.md
rb3ckers: Like I say in the other comment (that I was still typing) probably not. But let's have a chat about this also with Mark Bakker to get this sorted. Even if it is just to let him know we (again) hit limitations here

use/metrics-events/golden_signals.md
julianosk: Is it better now?

use/metrics-events/golden_signals.md
julianosk: @rb3ckers  Do you have an idea of how I could refrase this?

gggina (approved):

PR #525, opened 2021-07-01
DO NOT MERGE!! Auto propagation (v4.5)

konstantintatarnikov:
For v4.4, docs are updated to rename transparent > auto propagation.
This PR is to be reviewed and updated for v4.5

- Transparent propagation has been abstracted into Auto propagation which can be set `Auto` or `Transparent` by default
- `Auto` stands for Auto Propagation - new propagation function
- `Transparent` is what was earlier Transparent and it remains a default Auto propagation
- This goes to 4.4.0


fvlankvelt:
should the "Propagation functions" page be split up?  E.g. by explaining health propagation (at a high level) under "Use -> Health state" or "Configure -> Health".
The explanation of autoState "The precomputed auto state if returned from the script will lead to auto propagation" is a bit wordy.  Maybe "The state computed by the Auto Propagation function"?  It is also good to mention that the function is again invoked when the autoState changes.

Though little has changed here, should the "custom propagation function" mention that we assume idempotency?

gggina:
@fvlankvelt in the v4.4 docs there is now info on propagation in the use > health state section:
https://stackstate.gitbook.io/stackstate-docs-development/use/health-state/health-state-in-stackstate#propagated-health-state

PR #553, opened 2021-07-16
Stac 44 agent update

gggina:


gggina:
- separate page for openshift
- separate page for advanced configuration

PR #729, opened 2021-11-12
Fixed AWS zendesk links

hruhek:
https://stackstate.atlassian.net/browse/STAC-14831


gggina:
will merge to 4.5 docs first

PR #824, opened 2021-12-29
[v4.6] STAC-14858 and STAC-14860 rhs section renames and reordering

gggina:


gggina:
will merge 4.6 content directly to 4.6 branch

PR #865, opened 2022-01-27
Stac 15319 openmetrics

gggina:


stackpacks/integrations/openmetrics.md
```diff
@@ -0,0 +1,139 @@
...
+
+### Configuration
+
+To enable the OpenMetrics integration and begin collecting metrics data from an OpenMetrics endpoint, the OpenMetrics check must be configured on StackState Agent V2. The check configuration provides all details required for the Agent to connect to your OpenMetrics endpoint and retrieve the available metrics.
+
+{% tabs %}
+{% tab title="Kubernetes amd OpenShift" %}
+
+1. Deploy the Agent on your Kubernetes or OpenShift cluster.
+2. Add the following annotations when launching a Prometheus pod. This will collect all available metrics. It is also possible to specify a list of metrics to be fetched under `"metrics"`. This should either bea string representing the metric name or a mapping can be used to rename the metric`<EXPOSED_METRIC>:<SENT_METRIC>`:
```
rb3ckers, line 27: It's not about the Prometheus pod  (I admit that is confusing in the naming of the `prometheus_url`) but about the pod and container that exposes metrics via an openmetrics endpoint. So I would change it to:

> Add the following annotations when launching a pod that exposes metrics via an openmetrics endpoint

rb3ckers: Also a typo "bea" -> "be a"

stackpacks/integrations/openmetrics.md
```diff
@@ -0,0 +1,139 @@
...
+       ad.stackstate.com/api.instances: |
+        `[ 
+          {
+            "prometheus_url": "http://%%host%%:9404/metrics",
+            "namespace": "stackstate", 
+            "metrics": ["*"] 
+          } 
+        ]'
+   ...
+   ```
```
rb3ckers, line 43: In the example are a few things "hardcoded" that need to be changed by the user as well for their specific pod and application I'll put the parameterized example here:

```yaml
   ...
   metadata:
     annotations:
       ad.stackstate.com/<CONTAINER_NAME>.check_names: '["openmetrics"]'
       ad.stackstate.com/<CONTAINER_NAME>.init_configs: '[{}]'
       ad.stackstate.com/<CONTAINER_NAME>.instances: |
        `[ 
          {
            "prometheus_url": "http://%%host%%:<METRICS_PORT>/<METRICS_PATH>",
            "namespace": "<METRICS_NAMESPACE>", 
            "metrics": ["*"] 
          } 
        ]'
   ...
   # This already exists in the pod spec, the container name needs to match the container that is exposing the openmetrics endpoint
   spec:
     containers:
      - name: <CONTAINER_NAME>
   ```

With:
* CONTAINER_NAME: The name of the container that is exposing the openmetrics (there could even be multiple in a single pod that you all want to process (that's why there is a list in the json)
* METRICS_PORT: The port at which the openmetrics endpoint is exposed
* METRICS_PATH:  The path for the openmetrics api, often just `metrics`
* METRICS_NAMESPACE: All metrics collected here will get this as a dot-separated prefix

Comment (rb3ckers) [reply stackpacks/integrations/openmetrics.md:43]: You also can configure ALL the settings here that are available via the configuration file based option in the Linux/Windows tab so you could refer to that as well for more info on configuration options.

File: stackpacks/integrations/openmetrics.md [hunk 3]
```diff
@@ -0,0 +1,139 @@
...
+   - **extra_headers** - a list of additional HTTP headers to send in queries to the OpenMetrics endpoint. Can be combined with autodiscovery template variables. For example, `"Authorization: Bearer %%env_TOKEN%%"`.
+3. [Restart the StackState Agent\(s\)](../../setup/agent/about-stackstate-agent.md#deploy-and-run-stackstate-agent-v2) to apply the configuration changes.
+4. Once the Agent has restarted, wait for the Agent to collect data from the OpenMetrics endpoint and send it to StackState.
+
+{% endtab %}
+{% endtabs %}
+
+### Validation
+
+Run the Agent's status subcommand and look for `openmetrics` under the `Checks` section.
```
rb3ckers, line 119: This is pretty differnt for k8s and for linux/windows deployments. For linux (and I think also windows) you can just log into the machine and run `agent status`.

For kubectl it is a lot trickier:
* Lookup the node on which the pod you annotated is running
* Lookup the agent running on that node
* Run `kubectl exec <agent-pod-name> -n <agent-namespace> -- agent status`

This however applies to all agent checks running in Kubernetes so maybe better documented there and linked from here.

stackpacks/integrations/openmetrics.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Data collected
+
+### Metrics
+
+All metrics are retrieved from the specified OpenMetrics endpoint and available in the `StackState multi metrics` data source. Retrieved metrics will not automatically be mapped to topology elements.
```
rb3ckers, line 125: We also have a maximum of 2000 metrics retrieved. If you hit that you need to add filtering to grab the important metrics and still stay within the 2000.

rb3ckers (changes requested):

stackpacks/integrations/openmetrics.md
gggina: done

stackpacks/integrations/openmetrics.md
gggina: done

stackpacks/integrations/openmetrics.md
gggina: done

stackpacks/integrations/openmetrics.md
gggina: done

PR #887, opened 2022-02-17
add scale down/up instructions

gggina:


setup/data-management/backup_restore/kubernetes_backup.md
```diff
@@ -355,29 +357,44 @@ The indices restored are listed in the output, as well as the number of failed a
...
-3. The output should be:
+4. The output should be:
 
    ```javascript
    {
    "acknowledged" : true
    }
    ```
 
+5. After all indices have been deleted, scale up the `*2es` deployments again:
```
rb3ckers, line 394: I think it's better to first to the backup restore and only after that finished scale back up. Otherwise the services will create todays index again and if that is (partially) in the backup it will still result in issues

gggina: moved

PR #942, opened 2022-03-18
Stac 15432 add storage reqs for cops

rfosterstackstate:


setup/install-stackstate/requirements.md
```diff
@@ -79,6 +79,18 @@ If it is necessary to set a ResourceQuota for your implementation, the namespace
 
 KOTS requires VM running a [supported OS \(kurl.sh\)](https://kurl.sh/docs/install-with-kurl/system-requirements)
 
+### Disk Partitioning & latency requirements 
+
+For KOTS deployments you need to make sure there is at least 80GB allocate to the root (`/`) partition. Further to this it is advised you should allocate at least 500GB to `/var`.
```
vpartington, line 84: The 512 Gi should go to `/var/lib/longhorn`. The rest of `/var` can be on the root partition and will use up that 80 Gi.

vpartington: Also, I would make it a bullet list. They're easier to scan then proza.

vpartington (changes requested):

setup/install-stackstate/requirements.md
```diff
@@ -79,6 +79,19 @@ If it is necessary to set a ResourceQuota for your implementation, the namespace
...
+
+* `/` - at least 80GB
+* `/var/lib/longhorn` - at least 500GB 
+
+The `/var/lib/longhorn` disk should also have a latency of < 10ms. The following example cloud VM instance/disk combinations are known to provide sufficient performance for etcd and will pass the write latency preflight.
+
+* AWS m4.xlarge with 80 GB standard EBS root device
+* Azure D4ds_v4 with 80 GB ultra disk mounted at /var/lib/etcd provisioned with 2400 IOPS and 128 MB/s throughput
+* Google Cloud Platform n1-standard-4 with 500 GB pd-standard boot disk
+
```
gggina, line 94: For KOTS deployments, disks should be partitioned as follows:

* `/` - at least 80GB
* `/var/lib/longhorn` - at least 500GB with a latency of < 10ms

The following example cloud VM instance/disk combinations are known to provide sufficient performance for etcd and will pass the write latency preflight check.

* AWS m4.xlarge with 80 GB standard EBS root device
* Azure D4ds_v4 with 80 GB ultra disk mounted at `/var/lib/etcd` provisioned with 2400 IOPS and 128 MB/s throughput
* Google Cloud Platform n1-standard-4 with 500 GB pd-standard boot disk

PR #997, opened 2022-05-16
add sourceProperties to topologies JSON example

gggina:


configure/topology/send-topology-data.md
```diff
@@ -139,6 +137,7 @@ The JSON contains the following fields:
   * **externalId**: A unique ID for this component. This has to be unique for this instance.
   * **type**: A named parameter for this type.
   * **data**: A JSON blob of arbitrary data.
+  * **sourceProperties**: Optional. A JSON blob of arbitrary data. When populated, the contents of this field will be displayed in the StackState UI component properties in place of the `data` field.
```
Idorobots, line 140: Maybe worth mentioning, that regardless if `sourceProperties` is specified or not, `data` is still accessible in templates and the various functions that make use of this data?

Idorobots (approved):
Looks good to me :)

configure/topology/send-topology-data.md
gggina: added

PR #1000, opened 2022-05-17
[v5.0] Stac 16551 aad tuning

fvlankvelt:
introduce a new section "Anomaly detection" under "Configure", to host details on how to export feedback, scale up/down and troubleshoot using the status UI.
Draft site: https://stackstate.gitbook.io/stackstate-docs-development/v/stac-16551-aad-tuning/


configure/anomaly-detection/scaling.md
```diff
@@ -0,0 +1,58 @@
...
+The streams with an anomaly check have the highest priority in the AAD.  All anomaly health checks are updated on time when the number of **Checked streams** is higher than the number of **Streams with anomaly checks**.
+
+We can get these metrics from StackState using the CLI:
+
+```text
+sts metric get <multi_metrics_id> -t stackstate.spotlight_streams_with_anomaly_check -f anomaly_health_checks.csv
+sts metric get <multi_metrics_id> -t stackstate.spotlight_streams_checked -f checked.csv
+```
+
+Here `<multi_metrics_id>` is the ID for the "StackState Multi Metrics" datasource (`sts datasource list`).
```
jcshoekstra, line 47: should we script this too? such that running the full thing in one command is possible?

configure/anomaly-detection/status_ui.md
```diff
@@ -0,0 +1,68 @@
+---
+description: StackState Self-hosted v5.0.x
+---
+
+# Autonomous Anomaly Detector Status UI
```
jcshoekstra, line 5: not clear to me if we want to document this, and if we do, which (review) gates need to be passed

jcshoekstra: or if we should clearly state that it is a development / alpha portal

jcshoekstra: ah I realize now this documentation already existed, and this is simply moving it

configure/anomaly-detection/scaling.md
fvlankvelt: AFAICS it is not possible to extract the ID programmatically (i.e. with a one-liner).  The cli produces a table of all datasources, from which the ID can be copied.

PR #1182, opened 2022-09-28
STAC-17326: Document StackGraph restore default behaviour

aacevedoosorio:


setup/data-management/backup_restore/kubernetes_backup.md
```diff
@@ -264,17 +264,23 @@ Lines in the output that start with `Error from server (BadRequest):` are expect
 ### Restore a StackGraph backup
 
 {% hint style="warning" %}
-**When a backup is restored, the existing data in the StackGraph database will be overwritten.**
-
+**By default we only restore a backup on a clean environment to avoid some unexpected existing data loss.**
+If you are completely sure that the existing data can be overwritten then you can `-force` the command.
```
gggina, line 268:
```
**To avoid the unexpected loss of existing data, a backup can only be restored on a clean environment by default.**
If you are completely sure that any existing data can be overwritten, you can override this safety feature by using the command `-force`.
```

setup/data-management/backup_restore/kubernetes_backup.md
```diff
@@ -264,17 +264,23 @@ Lines in the output that start with `Error from server (BadRequest):` are expect
...
 {% endhint %}
 
-To restore a StackGraph backup, select a backup name and pass it as the first parameter in the following command:
+To restore a StackGraph backup on a clean environment, select a backup name and pass it as the first parameter in the following command:
 
 ```bash
 ./restore/restore-stackgraph-backup.sh sts-backup-20210216-0300.graph
 ```
 
+To restore a StackGraph backup on an **environment with existing data**, select a backup name and pass it as the first parameter in the following command next to a second parameter `-force`:
```
gggina, line 278: should we add a note here that existing data will be overwritten?
e.g.

```
{% hint style="info" %}
**Note that existing data will be overwritten when the backup is restored.**

Only do this if you are completely sure that any existing data can be overwritten.
{% endhint %}
```

setup/data-management/backup_restore/linux_backup.md
```diff
@@ -37,18 +37,31 @@ StackState topology and configuration data are stored in StackGraph. To create a
 
 ### Restore
 
+{% hint style="warning" %}
+**By default we only restore a backup on a clean environment to avoid some unexpected existing data loss.**
```
gggina, line 41: change as above

aacevedoosorio:
Closed in a different Mr vs version 5.0 of the docs

PR #1185, opened 2022-09-29
[v51] update references to JDK to state 11

gggina:


setup/install-stackstate/requirements.md
```diff
@@ -102,11 +102,7 @@ One of the following operating systems running Java. Check also the specific req
 
 #### Java
 
-OpenJDK 8 **patch level 121** or later.
-
-{% hint style="info" %}
-StackState **does not work** with JDK versions 9 or higher at this time.
-{% endhint %}
+OpenJDK 11 **patch level 121** or later.
```
aacevedoosorio, line 105: We can drop the patch level 121 ... that was really jdk 8 specific

PR #1394, opened 2023-01-11
STAC-17014: Fix capitalisation

hierynomus:


gggina:
will be merged to master after merging to 5.1

PR #1419, opened 2023-08-10
WIP: Document metric bindings and metrics

rb3ckers:


SUMMARY.md
```diff
@@ -23,7 +23,10 @@
 
 ## 📈 Metrics
 
+* [Explore Metrics](use/metrics/k8sTs-explore-metrics.md)
 * [Advanced](use/metrics/k8s-advanced.md)
+  * [Adding custom metrics to component](use/metrics/k8s-add-metrics.md)
```
craffit, line 28: to 'a component' or 'components' i think

use/metrics/k8s-add-metrics.md
```diff
@@ -0,0 +1,271 @@
...
+2. [Write the topology query (STQL) to select the components](#determine-the-topology-query)
+3. [Write the PromQL query for the desired metric](#determine-the-promql-query)
+4. [Bind the correct time series to each component](#bind-the-correct-time-serie-to-each-component)
+5. [Create or update the metric binding in StackState](#create-or-update-the-metric-binding-in-stackstate)
+
+As an example the steps will add a metric binding for the `Replica counts`  of Kubernetes deployments. This is just an example, this metric binding already exists in StackState by default.
+
+## Write the outline of the metric binding
+
+Use this template to create your own metric binding, all the parts will be filled in completely after going through this guide.
```
craffit, line 32: it is unclear here where i should place this template. maybe its best to already guide the user to make a .yaml file with this scaffolding

craffit: it might also be good to bring the cli settings apply command a bit forward in the explanation, to have the user get into a development loop already

use/metrics/k8s-openmetrics.md
```diff
@@ -0,0 +1,86 @@
+---
+description: StackState Kubernetes Troubleshooting
+---
+
+# OpenMetrics
+
+## Overview
```
rb3ckers, line 7: This openmetrics file can be ignored for now. It will be done as a separate PR. Therefore it is also not linked from the SUMMARY.md

use/metrics/k8s-add-metrics.md
```diff
@@ -0,0 +1,271 @@
...
+name: Container restarts
+priority: MEDIUM
+queries:
+- alias: Restarts - ${container_name}
+  expression: max by (kube_cluster_name, kube_namespace, pod_name, container_name) (kubernetes_containers_restarts{kube_cluster_name="${tags.cluster-name}", kube_namespace="${tags.namespace}", pod_name="${name}"})
+scope: (label = "stackpack:kubernetes" and type = "pod")
+unit: short
+```
+
+Note that the `alias` references the `container_name` label of the metric. Make sure the label is present on the query result, when the label is missing the `${container_name}` will be rendered as literal text to help troubleshooting.
```
craffit, line 271: suggestion: add a 'common problems' sections, with suggestions on how to debug issues around:
- metric binding not showing on component
- no data in metric chart
we could also do this later

use/metrics/k8s-add-metrics.md
rb3ckers: I had that and threw it out because there was too much information crammed that was not directly related to the task (creating a chart). I can try again...

rb3ckers: I'll try to have them make the file and won't explain everything initially. That seems to be a common approach in docs in general I think.

PR #1437, opened 2023-11-20
Stac 20194 doc aggregated topology monitors

LukaszMarchewka:


aacevedoosorio (approved):
I assume is the same text than in the description of the Monitor

LukaszMarchewka:
@aacevedoosorio yes the description is the same

PR #1451, opened 2024-03-05
Fixup ES restore scaling commands

whazenberg:
As the docs mentioned, some Deployments need to be scaled down before restoring from an ES snapshot. The scaling commands errored on a try on release 5.1.8. The Deployment names mismatched.

PR for changing the Deployment names in the docs. Assuming the default value for fullname used in the Chart.


hierynomus (approved):
Thanks Wytze!

PR #1455, opened 2024-03-15
STAC-21009 Updates related to Agent V3 and kubs_state_metrics.

LouisParkin:


fvlankvelt (commented):
looks like a global search and replace of "agent V2" with "agent V3".  If that's the case then it's fine by me.  Are you sure they need to go into the `5.1` branch?  Or should they go into the mainline `k8s-troubleshooter`?

LouisParkin:
> looks like a global search and replace of "agent V2" with "agent V3". If that's the case then it's fine by me. Are you sure they need to go into the `5.1` branch? Or should they go into the mainline `k8s-troubleshooter`?

@fvlankvelt it is not a "global find/replace" of V2 to V3 - There is still a distinction of sorts, the Agent V2 StackPack is still the Agent V2 StackPack, it is only the agent itself that has transitioned to V3.  All the internal versioning/messaging structures, URNs etc remain of the agent-v2 variety, as the change from agent v1 to v2 was a drastic change and required a separate stackpack, separate URNs, etc, the change from V2 to V3 was merely replacing the embedded python runtime in the container from python 2 to python 3.  So no, not a global replace, context matters, all references to the agent stackpack should still use V2.

Your second question, it should go into 5.1, as this is where the change occurred (Wytze asked about the discrepancy between docs and helm chart).  It is possible that the same needs to happen for v6, although that was not the point of STAC-21009.

fvlankvelt:
I must be confused then.  Why does the `5.1` release get the `V3` agent and is the `k8s-troubleshooting` mainline still on `V2` everywhere?

LouisParkin:
> I must be confused then. Why does the `5.1` release get the `V3` agent and is the `k8s-troubleshooting` mainline still on `V2` everywhere?

You are confused, indeed.  stackstate-k8s-agent has been on v3.0.0 for probably a year now...

fvlankvelt:
okay, I guess I had expected some references to that in the docs as well.

fvlankvelt (approved):

PR #1469, opened 2024-05-30
STAC-21328 Updated supported k8s versions.

LouisParkin:


setup/install-stackstate/requirements.md
```diff
@@ -13,10 +13,10 @@ Requirements for [StackState client \(browser\)](#client-browser) can be found a
 ### Supported versions
 
 StackState can be installed on a Kubernetes or OpenShift cluster using the Helm charts provided by StackState. These Helm charts require Helm v3.x to install and are supported on:
-* **Kubernetes:** 1.21 to 1.27
-* **Amazon Elastic Kubernetes Service (EKS):** 1.20 to 1.27
-* **Azure Kubernetes Service (AKS):** 1.20 to 1.26
-* **Google Kubernetes Engine (GKE):** 1.21 to 1.27
+* **Kubernetes:** 1.16 to 1.30
```
viliakov: We haven't tested Stackstate(platform) against newer versions of K8s.

LouisParkin: Have you tested stackstate on 1.27?

viliakov: You may say that. We run tenants in GKE 1.27.

setup/install-stackstate/requirements.md
```diff
@@ -13,10 +13,10 @@ Requirements for [StackState client \(browser\)](#client-browser) can be found a
 ### Supported versions
 
 StackState can be installed on a Kubernetes or OpenShift cluster using the Helm charts provided by StackState. These Helm charts require Helm v3.x to install and are supported on:
-* **Kubernetes:** 1.21 to 1.27
-* **Amazon Elastic Kubernetes Service (EKS):** 1.20 to 1.27
-* **Azure Kubernetes Service (AKS):** 1.20 to 1.26
-* **Google Kubernetes Engine (GKE):** 1.21 to 1.27
+* **Kubernetes:** 1.16 to 1.27
```
viliakov, line 16:
```suggestion
* **Kubernetes:** 1.21 to 1.27
```
Not sure if it was wrong copy/paste. Lets keep the lowest version as is

viliakov (approved):

setup/install-stackstate/requirements.md
LouisParkin: Hey @viliakov that is vanilla Kubernetes, so if someone was to try and run the agent on there, we are still fairly sure it would work, as all the breaking points (k8s 1.18, 1.21, 1.25, etc) were at some point all supported and tested at the same time, giving us confidence that the agent indeed still would run happily on such an old version.  I am going to leave it in fro now, and check back with Jeroen and Mark.

PR #1473, opened 2024-06-19
Stac 21222 troubleshooting is what we do

fvlankvelt:


README.md
```diff
@@ -1,10 +1,10 @@
-# StackState for Kubernetes troubleshooting
+# StackState
 
-**These documentation pages cover all functionality in the StackState for Kubernetes troubleshooting product**.
+**These documentation pages cover all functionality in the StackState product**.
```
janten, line 3: Not specific to this PR, but why do we never just say "... in StackState"?

janten (approved):

README.md
fvlankvelt: good point.  Maybe for us / the authors the distinction between the company and the product is significant.  But I agree that to readers there is only "StackState the product".

PR #1490, opened 2024-08-14
[DONOTMERGE] 7.x

fvlankvelt:
PR just for reviewing changes


classic.md
```diff
@@ -4,15 +4,15 @@ coverY: 0
...
+Select your version of Rancher Observability to jump to the associated documentation.
 
-### Kubernetes troubleshooting
+### Rancher Observability
 
-Use StackState to troubleshoot your Kubernetes cluster.
+Use Rancher Observability to troubleshoot your Rancher Kubernetes cluster.
 
-🚀 [StackState docs](https://docs.stackstate.com/)
+🚀 [Rancher Observability docs](https://docs.stackstate.com/)
```
rb3ckers: Shouldn't the self-hosted section get a new 6.0 entry pointing to the current main doc site (i.e. `k8s-troubleshooting` branch)?

rb3ckers (commented):
Looks like the logos (for StackState agent and StackState) are still the old logos.

use/logs/k8sTs-log-shipping.md
```diff
@@ -8,20 +8,20 @@ description: StackState v6.0
 
 ### Openshift
 
-Third-party log shippers are not readily supported on Openshift, the platform has options for log forwarding, which is used instead of the promtail configuration included in the StackState helm chart.  For detailed instructions on how to configure this, refer to the Kubernetes stackpack documentation on your running StackState instance. 
+Third-party log shippers are not readily supported on Openshift, the platform has options for log forwarding, which is used instead of the promtail configuration included in the Rancher Observability helm chart.  For detailed instructions on how to configure this, refer to the Kubernetes stackpack documentation on your running Rancher Observability instance.
```
rb3ckers: Do we even want to mention OpenShift here for the Rancher version?

fvlankvelt:
> Looks like the logos (for StackState agent and StackState) are still the old logos.

indeed.  The variants with text are not actually used.

classic.md
fvlankvelt: yeah probably.  I have a hard time coming up with a structure to organize the page.
Maybe the split should be on Prime/Community, putting the legacy SaaS docs under community?

use/logs/k8sTs-log-shipping.md
fvlankvelt: we still have lots of OpenShift references - they should be tackled in one fell swoop.  Might as well be this one, but I think Product should have the final say in this.

classic.md
rb3ckers: I'm not sure. Prime/community is not really the marketing that is used right now. So would just lead to confusion. Maybe check with Mark as well on Monday.

PR #1492, opened 2024-08-15
STAC-0: Recommend not using the upstream cluster

craffit:


k8s-suse-rancher-prime.md
```diff
@@ -60,14 +60,12 @@ This is an UI extension to Rancher Manager that integrates the health signals ob
 
 ### Where to install StackState server
 
-It is possible to install StackState server on the upstream cluster on which SUSE Rancher Prime is installed or it is possible to install StackState Server in a downstream cluster for Observability. See the below picture for reference.
-Our advice is to install StackState on a downstream cluster as shown on the right diagram.
-
+StackState server should be installed in a downstream cluster intended for Observability. See the below picture for reference.
```
hierynomus: Maybe highlight more importantly that the preference would be to run it in its _own_ downstream observability cluster, so not mixed with your production workloads.

rb3ckers (approved):

PR #1519, opened 2024-10-07
STAC-21796: Fix permissions for openshift

aacevedoosorio:


setup/install-stackstate/kubernetes_openshift/required_permissions.md
```diff
@@ -61,15 +61,15 @@ rules:
...
+  name: suse-observability-authentication
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
   name: system:auth-delegator
 subjects:
 - kind: ServiceAccount
-  name: stackstate-stackstate-k8s-api
-  namespace: stackstate
+  name: suse-observability-suse-observability-k8s-api
```
rb3ckers: Are you sure this is the correct name? I think it got duplicated because `stackstate` was the release and `stackstate-k8s` the helm chart. Now both are suse-observability (at least in the documented installation) so I would expect them to be deduplicated and without `k8s`, so `suse-observability-api`.

setup/install-stackstate/kubernetes_openshift/required_permissions.md
```diff
@@ -80,15 +80,15 @@ subjects:
...
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: stackstate-authorization
+  name: suse-observability-authorization
 subjects:
 - kind: ServiceAccount
-  name: stackstate-stackstate-k8s-api
-  namespace: stackstate
+  name: suse-observability-suse-observability-k8s-api
+  namespace: suse-observability
```
rb3ckers, line 91: Same question here.

rb3ckers (approved):

PR #1532, opened 2024-10-25
STAC-21693 Document migration from StackState to SUSE Observability

rb3ckers:


setup/data-management/backup_restore/configuration_backup.md
```diff
@@ -6,138 +6,74 @@ description: SUSE Observability Self-hosted
...
-{% hint style="info" %}
-Note that the [lock status](../../../stackpacks/about-stackpacks.md#locked-configuration-items) of configuration items installed by a StackPack configuration won't be included in the export.
-{% endhint %}
-
-To export configuration using the `sts` CLI, run the command:
-
-```text
-# Output in terminal window
-sts settings describe
+Scripts to work with configuration backups (but also all other backups)  can be found in the [restore directory of the SUSE Observability Helm chart repository \(github.com\)](https://github.com/StackVista/helm-charts/tree/master/stable/suse-observability/restore). To use the scripts, download them from GitHub or checkout the repository.
```
aacevedoosorio: Should we point the link to a specific version of the repo rather than master?

aacevedoosorio (dismissed):

setup/data-management/backup_restore/configuration_backup.md
rb3ckers: This is what we do in the backup documentation too. I think the reasoning was/is that we usually want to use the latest version, but I agree that is problematic if we already made changes. Pointing to a specific version means we'll need to update the docs for every change we make.

PR #1551, opened 2024-12-04
Added suse-observability_logs_collector script

amolkharche13:


configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
----
```
deontaljaard: Let's preserve the YAML frontmatter metadata variables.

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
...
-description: SUSE Observability Self-hosted v5.1.x
----
+# SUSE Observability logs-collector
 
-# Kubernetes logs
+## Notes
```
deontaljaard, line 12: To be explicit, could we add where the log is intended to collect logs from?

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
...
 
-## Overview
+Ensure you have `kubectl` access to the SUSE Observability cluster. If not, download the kubeconfig file and set it using the command: `export KUBECONFIG=$PATH-TO-YOUR/kubeconfig`.
 
-In a Kubernetes setup, SUSE Observability functions are distributed across different pods and logs for each function are stored per pod and container. You can access recent logs using `kubectl`, although for long term storage it's recommended to set up log aggregation.
+## Usage
 
-## Kubernetes pods for logging
+The script needs to be downloaded and run directly on the node, using the `root` user or `sudo`.
```
deontaljaard: The usage note seems slightly ambiguous compared to the earlier statement about access. Since the script can be run from any host where `kubectl` has access to the SO cluster, could we clarify this in the usage note?

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
...
 
-SUSE Observability logs are stored per pod and container. The table below shows the pod to access for logs relating to specific SUSE Observability functions. Note that actual pod names will include a number or random string suffix (for example, `stackstate-receiver-5b9d79db86-h2hkz`\) and may also include the release name specified when SUSE Observability was deployed as a prefix.
+### Download and run the script
+* Save the script as: `suse-observability_logs_collector.sh`
```
deontaljaard, line 21: Could we add the script to a `scripts` folder in the repo?

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
...
-
-* Shipping logs with [Fluentd \(fluentd.org\)](https://docs.fluentd.org/container-deployment/kubernetes)
-* A complete overview of setting up [log aggregation into Elasticsearch \(bitnami.com\)](https://docs.bitnami.com/tutorials/integrate-logging-kubernetes-kibana-elasticsearch-fluentd/)
-
-## See also
-
-* [kubectl command reference \(kubernetes.io/docs\)](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands)
+  Using `wget`:
+    ```bash
+    wget https://raw.githubusercontent.com/StackVista/stackstate-docs/SOME_PUBLIC_URL_TO_DOWNLOAD_SCRIPT/suse-observability_logs_collector.sh
```
deontaljaard: For now, the SOME_PUBLIC_URL_TO_DOWNLOAD_SCRIPT would be `refs/heads/suse-observability/scripts` - the full URL being: `https://raw.githubusercontent.com/StackVista/stackstate-docs/refs/heads/suse-observability/scripts/suse-observability_logs_collector.sh`

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
----
-description: SUSE Observability Self-hosted v5.1.x
----
+# SUSE Observability logs-collector
```
craffit: Could we add more of an introduction here, assuming a random user ends here with no context (what is the goal of this page in what context?)

configure/logging/kubernetes-logs.md
```diff
@@ -1,85 +1,27 @@
...
 
-# Kubernetes logs
+## Notes
 
-## Overview
+Ensure you have `kubectl` access to the SUSE Observability cluster. If not, download the kubeconfig file and set it using the command: `export KUBECONFIG=$PATH-TO-YOUR/kubeconfig`.
```
craffit: do we have some kind of error handling in the script in case kubectl is not on the path?

configure/logging/kubernetes-logs.md
amolkharche13: Please let me know if the following works or not.
"The script should be executed from a host that has kubectl configured with access to the SUSE Observability cluster. This can be done directly on the node using the root user or via sudo, or from any other host with the necessary permissions and connectivity to the cluster."

configure/logging/kubernetes-logs.md
amolkharche13: _scripts_ folder does not exist, Do you want me to create it and place the script under _scripts_ folder ?

configure/logging/kubernetes-logs.md
amolkharche13: We can add __if_ condition_ to check if KUBECONFIG set or not: If not set then exit script. I will add it.
```
if [[ -z "$KUBECONFIG" || ! -f "$KUBECONFIG" ]]; then
  echo "Error: KUBECONFIG is not set"
  exit 1
fi
```

configure/logging/kubernetes-logs.md
amolkharche13: Please let me know if the following works or not. 
Will add the below texts under overview  https://docs.stackstate.com/self-hosted-setup/install-stackstate/troubleshooting/kubernetes-logs#overview 
_"
This page provides tools and guidance for collecting logs from a SUSE Observability cluster. These logs are essential for diagnosing issues, identifying root causes, and troubleshooting problems in clusters that may not be functioning as expected."_

configure/logging/kubernetes-logs.md
deontaljaard: That works!

configure/logging/kubernetes-logs.md
deontaljaard: Yes, let's include the script in this PR.

amolkharche13: Added

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
deontaljaard, line 1: Nice script. I tested it and it worked like a charm. 

I do have a couple of suggestions, though...

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
...
+# suse observability 1.x logs collector
+
+# Namespace to collect information
+NAMESPACE="suse-observability"
```
deontaljaard, line 4: A few thoughts regarding the namespace:

- I think it will be a nice add to be able to pass the namespace as an argument. 
- Also, to validate whether the namespace exists - if not, then exit.
- In order to do the validation, it'll also be useful to check whether `kubectl` is installed and accessible (on the path - like @craffit asked somewhere else).

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
...
+
+# Check if KUBECONFIG is set
+if [[ -z "$KUBECONFIG" || ! -f "$KUBECONFIG" ]]; then
+  echo "Error: KUBECONFIG is not set"
```
deontaljaard: I think it'll be nice to provide specific guidance on how to set the KUBECONFIG env var properly.

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo "Please ensure KUBECONFIG is set to the path of a valid kubeconfig file before running this script. Exiting..."
```
deontaljaard: I like what the [rancher collector script is doing with echos](https://github.com/rancherlabs/support-tools/blob/master/collection/rancher/v2.x/logs-collector/rancher2_logs_collector.sh#L1270-L1274) - annotating log messages with timestamps and preserving the output of a script in a log file. Shall we do the same?

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
...
+echo "Collecting events in $NAMESPACE ..."
+kubectl -n "$NAMESPACE" get events --sort-by='.metadata.creationTimestamp' > "$OUTPUT_DIR/events"
+
+archive_and_cleanup() {
+    echo "Creating archive $ARCHIVE_FILE..."
+    tar -czf "$ARCHIVE_FILE" "$OUTPUT_DIR"
+    echo "Archive created."
+
+    echo "Cleaning up the output directory..."
+    rm -rf "$OUTPUT_DIR"
```
deontaljaard, line 112: Can we make this clean up optional - by setting some for example? I ran the script locally, but had to unpack the tarball again anyway.

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,105 @@
...
+    kubectl -n "$NAMESPACE" get cronjob -o yaml > "$OUTPUT_DIR/yaml/cronjob.yaml"
+
+    echo "YAML configurations collected."
+}
+
+# Function to collect pod logs
+collect_pod_logs() {
+    echo "Collecting pod logs..."
+    PODS=$(kubectl -n "$NAMESPACE" get pods -o jsonpath="{.items[*].metadata.name}")
+    for pod in $PODS; do
```
deontaljaard, line 65: I few thoughts on the pod log collection:
- I noticed for one cluster the pod log collection taking some time - it would be nice to in a future version of the script to make pod log collection execute in parallel.
- Supporting time ranges in the log collection commands could be useful for scenarios where one is only interested in a specific period, which could speed up the log collection as well. Supporting the time range flags could be added in a future version of the script.

deontaljaard (commented):
Looking good! Added a few remarks to the script.

scripts/suse-observability_logs_collector.sh
amolkharche13: We already added in [Notes ](https://github.com/StackVista/stackstate-docs/blob/06882e82801b3b3ee2327ea6fe5231daeaf14ebd/configure/logging/kubernetes-logs.md#notes) section. 
Added in output as well.

scripts/suse-observability_logs_collector.sh
amolkharche13: Sometimes, the SO cluster manages over 100+ nodes, which can result in a large number of files. By using the `tar` command, we can reduce space usage by up to 90%. Also, its easy to upload to the case as well.

scripts/suse-observability_logs_collector.sh
amolkharche13: Will take a note of it.

scripts/suse-observability_logs_collector.sh
amolkharche13: By default, we consumed "suse-observability" ns. Is SO cluster can be installed in other ns as well and works fine without any dependency ?
I would suggest making it defaults with suse-observability ns. 

Yes, we can validate whether the namespace exists - if not, then exit.

scripts/suse-observability_logs_collector.sh
amolkharche13: This refers to timestamp command and it does not installed by default. Will do it using _date_ command.

DONE

scripts/suse-observability_logs_collector.sh
amolkharche13: Please check now, It won't take much time. 
```
root@jumpsrv:~/terraform/suse-observability# bash suse-observability_logs_collector_v1.sh
2024-12-05 12:39:31: Collecting node details...
2024-12-05 12:39:39: Collecting pod statuses...
2024-12-05 12:39:41: Collecting StatefulSets information...
2024-12-05 12:39:56: Collecting DaemonSets information...
2024-12-05 12:40:02: Collecting Deployments information...
2024-12-05 12:40:04: Collecting services information...
2024-12-05 12:40:06: Collecting information about configmaps and secrets...
2024-12-05 12:40:09: Collecting cronjob information...
2024-12-05 12:40:11: Collecting events in suse-observability ...
2024-12-05 12:40:12: Collecting pod logs...
2024-12-05 12:40:34: Collecting YAML configurations...
Creating archive suse-observability_logs_20241205123931.tar.gz...
Archive created.
Cleaning up the output directory...
Output directory removed.
All information collected in the suse-observability_logs_20241205123931.tar.gz
root@jumpsrv:~/terraform/suse-observability#
```

scripts/suse-observability_logs_collector.sh
deontaljaard: Yep, I get that. My suggestion about making it optional was meant to say that I would like to enable/disable the clean up of the output dir, and eventcreating a tarball. If I'm running the script locally just for myself, I don't want to clean up the output dir (and create a tarball for that matter). 

However, the current implementation is also fine as is (to me) - I can just extract the tarball.

scripts/suse-observability_logs_collector.sh
deontaljaard: Agreed. The default namespace will stay `suse-observability`, yes. 

But you can deploy suse-observability to another namespace, and my suggestion was to cater for that.

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,122 @@
...
+    techo "Collecting pod logs..."
+    PODS=$(kubectl -n "$NAMESPACE" get pods -o jsonpath="{.items[*].metadata.name}")
+    for pod in $PODS; do (
+        mkdir -p "$OUTPUT_DIR/pods/$pod"
+        CONTAINERS=$(kubectl -n "$NAMESPACE" get pod "$pod" -o jsonpath="{.spec.containers[*].name}")
+        for container in $CONTAINERS; do
+            kubectl -n "$NAMESPACE" logs "$pod" -c "$container" > "$OUTPUT_DIR/pods/$pod/${container}.log" 2>&1
+            kubectl -n "$NAMESPACE" logs "$pod" -c "$container" --previous > "$OUTPUT_DIR/pods/$pod/${container}_previous.log" 2>/dev/null
+        done
+        ) &
```
deontaljaard: Using background processes for concurrency is a neat approach,  but spawning an unbounded number of subprocesses could lead to excessive system resource usage, such as high CPU or memory consumption.

To address this, consider implementing a mechanism to cap the number of subprocesses running concurrently. This could be achieved using a semaphore-like construct or a worker pool approach.

amolkharche13: I have no idea how to use semaphore in bash and never heard before. I will learn and implement it in our script. 
By the way how many concurrent processes we need to set e.g. 2 or 5 or 10 or 15 etc.?

Is it fine if we can implement semaphore in next version of script? For now, we can publish script without background processes, I will revert the changes.

deontaljaard: Good question. I'd start small, observe and tweak accordingly, trying not to exceed some limit - but I don't have a lot of knowledge about bash subprocesses. I'd be okay with leaving the performance optimisation for some future version - we can observe if it really needs the optimisation before adding the complexity (if at all).

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,122 @@
...
+    # DaemonSet YAMLs
+    kubectl -n "$NAMESPACE" get daemonsets -o yaml > "$OUTPUT_DIR/yaml/daemonsets.yaml"
+    # Service YAMLs
+    kubectl -n "$NAMESPACE" get services -o yaml > "$OUTPUT_DIR/yaml/services.yaml"
+    # Deployment YAMLs
+    kubectl -n "$NAMESPACE" get deployments -o yaml > "$OUTPUT_DIR/yaml/deployments.yaml"
+    # ConfigMap YAMLs
+    kubectl -n "$NAMESPACE" get configmaps -o yaml > "$OUTPUT_DIR/yaml/configmaps.yaml"
+    # Secret YAMLs
+    kubectl -n "$NAMESPACE" get secrets -o yaml > "$OUTPUT_DIR/yaml/secrets.yaml"
```
deontaljaard: I'm not sure we need the secrets? I'm concerned it will leak sensitive information.

amolkharche13: Good point, that makes sense to me. 
Yes, it can indeed pose a security risk. If we don't need Kubernetes secrets for troubleshooting, it's best to avoid requesting them from the customer to prevent exposing sensitive information.
I will modify the script.

scripts/suse-observability_logs_collector.sh
amolkharche13: In the coming weeks, I plan to enhance the next version of the script:  

```
1. Schedule the script to run via cron every 15 minutes, four consecutive times (currently using `&`, which spawns an unbounded number of subprocesses).  
2. Test the script across various environments like Rancher, EKS, and DigitalOcean.  
3. Evaluate resource utilization, including CPU, memory, and other performance metrics, during execution.  
4. Assess whether optimization is necessary, as introducing semaphores could add complexity to the code.  
``` 

What are your thoughts?

deontaljaard: That sounds like a thorough approach to determine what performance and stability optimisations could be made.

I think we can wait until it's really necessary, though - and basing follow-up improvements on feedback that we get from users of the script. However, if you have the capacity/time to spend time on this, feel free to do so - but maybe there are other tasks that you want to work on as well.

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,118 @@
...
+    # DaemonSet YAMLs
+    kubectl -n "$NAMESPACE" get daemonsets -o yaml > "$OUTPUT_DIR/yaml/daemonsets.yaml"
+    # Service YAMLs
+    kubectl -n "$NAMESPACE" get services -o yaml > "$OUTPUT_DIR/yaml/services.yaml"
+    # Deployment YAMLs
+    kubectl -n "$NAMESPACE" get deployments -o yaml > "$OUTPUT_DIR/yaml/deployments.yaml"
+    # ConfigMap YAMLs
+    kubectl -n "$NAMESPACE" get configmaps -o yaml > "$OUTPUT_DIR/yaml/configmaps.yaml"
+    # Cronjob YAMLs
+    kubectl -n "$NAMESPACE" get cronjob -o yaml > "$OUTPUT_DIR/yaml/cronjob.yaml"
```
rb3ckers, line 57: Can we also add the pods to this list? Most important part is that the yaml of a pod contains all details on pod status.

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,118 @@
...
+kubectl -n "$NAMESPACE" get pods -o wide > "$OUTPUT_DIR/pods_status"
+
+# Collect StatefulSets information
+techo "Collecting StatefulSets information..."
+kubectl -n "$NAMESPACE" get statefulsets -o wide > "$OUTPUT_DIR/statefulsets"
+kubectl -n "$NAMESPACE" describe statefulsets > "$OUTPUT_DIR/statefulsets_describe"
+
+# Collect DaemonSets information
+techo "Collecting DaemonSets information..."
+kubectl -n "$NAMESPACE" get daemonsets  > "$OUTPUT_DIR/daemonsets"
```
rb3ckers: Is this on purpose without `-o wide`?

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,118 @@
...
+            kubectl -n "$NAMESPACE" logs "$pod" -c "$container" > "$OUTPUT_DIR/pods/$pod/${container}.log" 2>&1
+            kubectl -n "$NAMESPACE" logs "$pod" -c "$container" --previous > "$OUTPUT_DIR/pods/$pod/${container}_previous.log" 2>/dev/null
+        done
+    done
+ }
+
+
+# Collect general pod statuses
+techo "Collecting pod statuses..."
+kubectl -n "$NAMESPACE" get pods -o wide > "$OUTPUT_DIR/pods_status"
```
rb3ckers, line 78: No describe for pods? If we do the `get -o yaml` I think it won't be necessary, but the describe is here for the other resource types.

scripts/suse-observability_logs_collector.sh
amolkharche13: Added

scripts/suse-observability_logs_collector.sh
amolkharche13: Done

scripts/suse-observability_logs_collector.sh
amolkharche13: Added  get -o yaml for pod

scripts/suse-observability_logs_collector.sh
amolkharche13: I believe no daemonsets are deployed during the initial cluster installation. They are only installed after the agent ,which is why they were not included initially. 
But I have added in script.

configure/logging/kubernetes-logs.md
rb3ckers: I still don't get this. Why would the script need to run as root user or via sudo and what `node` is this referring to?

I would expect to be able to run this script as whatever user on any host that has kubectl and a kubeconfig that allows access to the cluster with sufficient read permissions to read the data.

rb3ckers: Oh I see there is also mention of `the node` and the root /sudo requirement on line 14. But I still don't get which node that is or why it would need to run as root.

amolkharche13: Node means any node where you have kubectl access to the SO cluster.

amolkharche13: We can remove _usage_ content line if thats not needed.
```
## Usage
The script needs to be downloaded and run directly on the node, using the `root` user or `sudo`.
```

rb3ckers: Aha ok. Now I understand, but having the permissions to write files to the local file system is a much weaker permission than `root` or `sudo` access. The latter many users will not have, but almost any user will be able to write to `/tmp` for example. So I would remove the "scary" references to needing root access and replace them with the requirement to have permission to write to the file system.

I think my confusion on the `node` comes from the fact that in the first line it is called `host`:
> The script should be executed from a host

While after that it is referred to as `node`, which then I associate with Kubernetes nodes and not related to the previously mentioned `host`. Can we just use `host` everywhere to be consistent?

amolkharche13: Added host keywords

scripts/suse-observability_logs_collector.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
+# suse observability 1.x logs collector
```
deontaljaard: What does the version in the comment refer to? Initially, I thought it was referring to the version of SUSE Observability, but it is no longer at this version, which can lead to confusion at first glance.  

I think we should remove it.

deontaljaard (dismissed):
One minor remark about a comment in the script, otherwise LGTM. 

I'd still wait for an approval from @craffit or @rb3ckers as well before merging.

scripts/suse-observability_logs_collector.sh
amolkharche13: Done

deontaljaard:
@amolkharche13, JFYI, there are still conflicts to resolve.

deontaljaard (approved):
LGTM.

Like in my previous approval, I'd wait for either @craffit and @rb3ckers to approve as well (they also still have open conversations - which do seem resolved).

craffit (approved):

PR #1576, opened 2025-01-27
STAC-22297 Added documentation for auto-instrumentation of AWS Lambda…

LouisParkin:
…s with OTEL.


setup/otel/languages/node.js/auto-instrumentation-of-lambdas.md
```diff
@@ -0,0 +1,152 @@
...
+- **`OTLP_API_KEY`:** Authenticates your Lambda function to send data to SUSE Observability.
+- **`OTLP_ENDPOINT`:** Specifies the address of your SUSE Observability instance.
+- **`OPENTELEMETRY_COLLECTOR_CONFIG_FILE`:** Points to the configuration file for the OpenTelemetry Collector.
+- **`AWS_LAMBDA_EXEC_WRAPPER`:** Configures the Lambda execution environment to use the OpenTelemetry handler.
+- **`OTLP_INSTR_LAYER_ARN`:** Provides the ARN (Amazon Resource Name) of the OpenTelemetry instrumentation layer, which adds the necessary components for auto-instrumentation.
+- **`OTLP_COLLECTOR_LAYER_ARN`:** Provides the ARN of the OpenTelemetry collector layer, which is responsible for receiving, processing, and exporting telemetry data.
+
+**Important Considerations:**
+
+- **GRPC Endpoint:** The `OTLP_ENDPOINT` should specify the gRPC endpoint of your SUSE Observability instance without any `http` or `https` prefix. Use port 443 for secure communication.
```
fvlankvelt, line 41: it's not clear to me why there is the intermediate collector between lambda and suse observability if the lambda is going to send data directly to suse observability anyway.

LouisParkin: Hi @fvlankvelt this is as a result of @markbakker78 requesting it to be so.  Apparently the default client deployment model advises a collector instance that aggregates / pre-processes data before sending it to StackState.  I argued that we don't have it in StackState and lambdas typically generate such low volumes, it is not needed, but his response was that, since we already advise customers to have it, it makes sense to send the data to that collector instance.

setup/otel/languages/node.js/auto-instrumentation-of-lambdas.md
```diff
@@ -0,0 +1,152 @@
...
+      receivers: [otlp]
+      exporters: [debug, otlp/stackstate]
+      processors: []
+    metrics:
+      receivers: [otlp]
+      exporters: [debug, otlp/stackstate]
+      processors: []
+```
+
+Be aware this collector is used to send the data over to a next collector which then is used for tail sampling, metric aggregation, etc. before sending data over to SUSE Observability. This second collector also needs to run in the customers environment. See this page for [instructions](../../collector.md).
```
fvlankvelt: I would imagine that collector can also add the authentication header for SUSE Observability?

setup/otel/languages/node.js/auto-instrumentation-of-lambdas.md
LouisParkin: But I also see how this conflicts with what is written in the docs.  I'll ask @markbakker78 how he would like to handle it.

setup/otel/languages/node.js/auto-instrumentation-of-lambdas.md
LouisParkin: You would imagine things I have not yet dared to imagine.

setup/otel/languages/node.js/auto-instrumentation-of-lambdas.md
fvlankvelt: I think the architecture is indeed fine as a default - but then that aggregator should also do the authentication with StackState.  How the lambda authenticates with the aggregator is up to the customer.

LouisParkin: @fvlankvelt  that is not part of this discussion though.  That is a separate topic. (even linked to in my document update).  If you are saying that the document linked to `../../collector.yaml` doesn't deal with this, then that is another update required I guess.

markbakker78 (approved):
After a discussion with Louis, Louis and me we decided to merge this first version and wait for feedback to get this going,

PR #1608, opened 2025-04-19
Update suse-observability_logs_collector.sh to accept 1st parameter a…

gtam:
Allow users to specify a namespace for log collection in case they deployed into a namespace not "suse-observability".


rb3ckers (approved):

rb3ckers:
Thanks @gtam

PR #1621, opened 2025-06-02
STAC-22208: document RBAC on Rancher

fvlankvelt:


setup/security/rbac/rbac_rancher.md
```diff
@@ -0,0 +1,171 @@
+---
+description: SUSE Observability Self-hosted
+---
+
+## Overview
+
+The SUSE Rancher Prime Observability Extension allows RBAC configuration of users access in SUSE Observability.
```
craffit: I would immediately note here that this uses k8s' rbac system to grant access to rancher users, and can be configured through the concepts in rancher (role templates).

I would also immediately point the user here to the altenrative rbac implemntation we have (legacy rbac?) to avoid confusion

setup/security/rbac/README.md
```diff
@@ -4,3 +4,29 @@ description: SUSE Observability Self-hosted
...
+RBAC is an authorization system that provides fine-grained access management of SUSE Observability resources, a clean and easy way to audit user privileges and to fix identified issues with access rights.
+
+## What can I do with RBAC?
+
+Here are some examples of what you can do with RBAC:
+
+* Allow one user to have access to the development cluster only, another one to both the production and development cluster and a third can access the development cluster and only 1 namespace in the production cluster.
+* Give a small group of users an administrator role to setup and configure SUSE Observability. While giving all developers a troubleshooter role to view all topology, metrics, logs and events, but with limited configuration capability.
+
+## What's a role in SUSE Observability?
```
craffit, line 22: this section pretty much sums up the different system we have, I'd be more clear in the title what we have two systems. Proposal: kubernetes rbac vs legacy rbac

setup/security/rbac/rbac_rancher.md
```diff
@@ -0,0 +1,171 @@
+---
```
craffit, line 1: We should adapt each 'legacy' page to point out we also have new rancher/kubernetes based rbac system

setup/security/authentication/oidc.md
```diff
@@ -19,8 +19,55 @@ Before you can configure SUSE Observability to authenticate using OIDC, you need
...
+  authentication:
+    rancher:
+      clientId: "<oidc-client-id>"
+      secret: "<oidc-secret>"
+      baseUrl: "<rancher-url>"
+```
+You can override and extend some of the OIDC config for Rancher with the following fields:
+- `discoveryUri`
+- `redirectUri`
+- `customParams`
```
aacevedoosorio: Do we want an example or field description?

setup/security/rbac/rbac_rancher.md
```diff
@@ -0,0 +1,171 @@
...
+description: SUSE Observability Self-hosted
+---
+
+## Overview
+
+The SUSE Rancher Prime Observability Extension allows RBAC configuration of users access in SUSE Observability.
+
+Two kinds of roles are used for accessing SUSE Observability:
+* A *scope role* (Observer) grants access to data - either all data in a SUSE Observability instance, data coming from a cluster, or just the data for a namespace.  This role is provisioned in a cluster to be observed.
+* An *instance role* grants permissions to access or modify functionality of SUSE Observability itself.
```
aacevedoosorio, line 12: Do we want to add that grants access to observed data from all clusters?

setup/security/rbac/rbac_rancher.md
```diff
@@ -0,0 +1,171 @@
...
+
+The observer role grants a user the permission to read topology, metrics, logs and trace data for a namespace or a cluster.  There are three `RoleTemplate`s that grant access to observability data:
+
+* **Observer** - grants access to data coming from namespaces in a Project.  This can be used in the "Project Membership" section of the cluster configuration.
+* **Cluster Observer** - grants access to all data coming from a Cluster.  This template can be used in the "Cluster Membership" section of the cluster configuration.
+* **Instance Observer** - grants access to all data in a SUSE Observability instance.  This template can be used on the Project that includes SUSE Observability itself.
+
+### Instance roles
+
+There are four roles predefined in SUSE Observability:
```
aacevedoosorio: We are missing the fourth (Observer)

aacevedoosorio (commented):
Perhaps we already have it (or not) but it would be great to have a note on the legacy/standalone rbac system about not being able to scope metrics and traces over there ...

setup/security/rbac/README.md
fvlankvelt: I think this was moved verbatim from the "legacy RBAC" docs there were before.

But agreed, it makes sense to make the difference clear from the start.

setup/security/authentication/oidc.md
fvlankvelt: added a description - the rancher OIDC just works the same as any other one (I hope)

setup/security/rbac/rbac_rancher.md
fvlankvelt: I've moved things a bit around - to make clear that the Admin & Troubleshooter get to see all the data and the Observers need the Recommended Access.

setup/security/rbac/rbac_rancher.md
fvlankvelt: I've shuffled it a bit.  This role was already present under the "Observers" - hopefully it's now a bit clearer.

PR #1625, opened 2025-06-12
STAC-21030: Explaining how to configure K8s affinities

viliakov:


setup/install-stackstate/kubernetes_openshift/affinity.md
```diff
@@ -0,0 +1,122 @@
+---
+description: SUSE Observability Affinity Configuration
```
rb3ckers, line 2: This is a new file. To get it published using Gitbook it needs to be included somewhere in the `SUMMARY.md`. But I guess, now that this needs to be moved to the new docs-site, things will be different but I would expect the file still needs to be added to the sidebar.

PR #1628, opened 2025-09-11
STAC-23332 Delete ElasticSearch indices before a restore operation.

LukaszMarchewka:


viliakov (approved):

setup/data-management/backup_restore/kubernetes_backup.md
```diff
@@ -646,7 +646,7 @@ To delete existing Elasticsearch indices so that a snapshot can be restored, fol
 ### Restore an Elasticsearch snapshot
 
 {% hint style="danger" %}
-**When a snapshot is restored, existing indices won't be overwritten.**
+**When a snapshot is restored, existing indices won't be overwritten, and the operation will fail. You have to delete the indices before restoring a snapshot. Be careful, this will delete all data in the indices, and the indices will be recreated with the data from the snapshot.**
```
viliakov:
```suggestion
**You must delete the target indices before restoring a snapshot. This will delete all existing data in those indices. The indices will then be recreated with the data from the snapshot. This step is necessary because existing indices cannot be overwritten during restoration, causing the operation to fail.**
```

I asked AI to rewrite this note and put an action first and provide the reasoning for it later. In my opinion it is a better approach.

viliakov (dismissed):

viliakov (approved):

akashraj4261 (approved):

